Issue #7, opened 2018-04-07
Add Ask.com Data Source


Ice3man543:

Mzack9999:
implemented with PR #27

PR #27, opened 2018-05-03
added ask search engine

Mzack9999:
Implement Issue #7

### DIFF
--- a/libsubfinder/engines/passive/passive.go
+++ b/libsubfinder/engines/passive/passive.go
@@ -35,6 +35,7 @@ import (
 	"github.com/Ice3man543/subfinder/libsubfinder/sources/virustotal"
 	"github.com/Ice3man543/subfinder/libsubfinder/sources/waybackarchive"
 	"github.com/Ice3man543/subfinder/libsubfinder/sources/baidu"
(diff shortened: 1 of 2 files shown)


Ice3man543:
Thanks a lot @Mzack9999 for this. Will look into it in 4-5 hours when I am free. Do you want me to add you as a collaborator? You are doing great work. Keep it up.

Mzack9999:
Sure, feel free to add me

Ice3man543:
This gives me 18 results for google.com. Is that fine @Mzack9999 @codingo

Ice3man543:
Done. Added you as a collaborator :+1: !

Mzack9999:
I think it behaves as expected (for example sublist3er gives zero results), in order to get more results probably a configurable parameter defining the minimum recursion level should be considered? The crawler stops after a minimum of 15 (for now it's hardcoded) if no new results are found

Ice3man543:
Thanks a lot. Will merge this once I return. I will add a --set-settings AskMinimum flag so that it can be configured. I will do something similar for all the awesome search engines you have added. :+1:

Ice3man543:
Added --set-config AskPages=100 flags in commit 98518d5782be09cd54b0a9e5fe2e38f5f4de9391
:+1:

PR #29, opened 2018-05-04
Added Bing passive search

Mzack9999:


Ice3man543:
Whoa man, you're on fire. How may times do I have to thank you :grinning: ! Thanks again.

Ice3man543:
Added --set-config BingPages=100 flags in commit 98518d5782be09cd54b0a9e5fe2e38f5f4de9391
:+1:

PR #42, opened 2018-05-10
Multiple domain list fix

### LINKED ISSUE #34: Multiple domain list input not working correctly
The -dl switch is not working correctly. It outputs subdomains for other domains that the one the scan Is being run for. 
This needs to be fixed.

Mzack9999:
- Changed flag dl to dL
- Implemented subdomain validity check (TODO: to be tested before merging, I did just some rapid test)
Fixes #34

### DIFF
--- a/libsubfinder/engines/passive/passive.go
+++ b/libsubfinder/engines/passive/passive.go
@@ -341,6 +341,7 @@ func PassiveDiscovery(state *helper.State) (finalPassiveSubdomains []string) {
 					fmt.Printf("\n[-] Searching For Subdomains in Bing")
 				}
 				sourceConfig.Bing = true
+				sourceConfig.NoOfSources = sourceConfig.NoOfSources + 1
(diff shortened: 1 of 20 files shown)


Mzack9999:
Sorry about the double flag renaming, I just noticed it now.
About the validation, if I didn't misunderstood, here the issue was about some invalid result being printed in real time as soon as they are discovered, so I introduced this check where it was missing as rapid temporary fix, and later on working on a refactoring of the whole passive section

Ice3man543:
@Mzack9999 We basically need a structure containing keys as domains loaded from list and values as an array []string containing subdomains found or something similar. A complete rewrite by you would be much appreciated and welcomed since the engine has became messy. Also, change the name of enumerate.

Mzack9999:
Ok @Ice3man543, things are clearer now, I will close then this PR and will open a new one with a rewrite of the passive engine

PR #44, opened 2018-05-11
fix for  < character

### LINKED ISSUE #39: Error: character '<' looking for beginning of value during scans (scan still completes)
## What's the problem (or question)?
included in the output is the following string:

   error: invalid character '<' looking for beginning of value

## How can we reproduce the issue?
1.  install golang 1.10.2 in a debian system
2. go get github.com/Ice3man543/subfinder
3. $GOPATH/bin/subfinder -d reddit.com

before the domains are output, the error message presents itself.

Mzack9999:
Fixes #39

### DIFF
--- a/libsubfinder/sources/riddler/riddler.go
+++ b/libsubfinder/sources/riddler/riddler.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"errors"
 
 	"github.com/Ice3man543/subfinder/libsubfinder/helper"
 )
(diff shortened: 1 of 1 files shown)


Mzack9999:
@Ice3man543 This was a fast "fix" in order to give at least the reason of failure. I suppose it's not necessary in others, I debugged the issue and it happens just in this module. I plan to change also every passive source to be totally async and reduce the usage of long living objects to minimum and meanwhile I will also fix these not working modules, I can't allocate a lot of time right now, so it will take some time

Ice3man543:
@Mzack9999 Awesome work. Really appreciate you contributing to this tool. No need to hurry. Just, create a branch and make one source asynchronous. Others I can do following what you do to one. What are those not working modules?

Mzack9999:
I saw this one, and there was another one that produced weird results (probably an issue related to regex or parsing), later on filtered out by the final validation. I said generically "not working modules" to mean that I will try to fix these issues along the way if it happens I work on such modules

PR #52, opened 2018-05-19
changed FROM iron/go:dev to golang for RPi support.

x1mdev:
Tested this on my Raspberry Pi 3 with Ubuntu 16.04 and it's working fine with this little change. 

Also tested it on my main OS (Fedora 27) and no issues occurred.

- x1m


Ice3man543:
Thanks for this. Merging it in a moment.

PR #53, opened 2018-05-25
Worker pool passive sources

Mzack9999:
- Applied worker pool to all passive sources (since it's used already in passive, resolver and soon bruteforcer)
- Error printed out directly in passive source module (@Ice3man543 I think this allows a more detailed output and flow control)
- Small refactors and code improvement

TODO: That's another significant change, I tested it and all seems ok, but please test it too before merging :)

### DIFF
--- a/libsubfinder/engines/passive/passive.go
+++ b/libsubfinder/engines/passive/passive.go
@@ -327,12 +327,12 @@ func discover(state *helper.State, domain string, sourceConfig *Source) (subdoma
 			}
 		}
 		results := job.Result.([]string)
-		for _, subdomain := range result.Subdomains {
+		for _, subdomain := range results {
 			finalPassiveSubdomains = append(finalPassiveSubdomains, subdomain)
 		}
 	}
 
-	passivePool.Stop()
+	domainDiscoverPool.Stop()
 
 	// Now remove duplicate items from the slice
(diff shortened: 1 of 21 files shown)


Ice3man543:
@Mzack9999 You're on fire man. I think your idea is right. Make things unified. Actually, i've already implemented the same idea in bruteforcer. Just some tests and everything will be fine.
One thing - can you please pull the recent changes and incorporate them since i've made some alphabatizing changes and others. Other than that :+1:

Ice3man543:
@Mzack9999 Also, your previous work was giving same line two time saying Output written to file. I traced this line back to **output** library, and i know it's been called twice. Can you check?

Ice3man543:
@Mzack9999 Also, i am pushing bruteforcer now with some more changes i made. Pull those and resolve them if you have time, if not tell me :)

Ice3man543:
@Mzack9999 I traced the problem and found it's source. I am going to rewrite the output engine now :)

Mzack9999:
@Ice3man543 I think I will have some time tomorrow, now I'm out :/

Ice3man543:
Ok. No problem. :)

Mzack9999:
@Ice3man543 I pulled changes from master, there are no conflicts and everything seems to work as expected

Issue #55, opened 2018-05-28
i have no idea how to download it 


samuel823:
u should make a video or something , ur explain is so bad , im sure no one can understand it .....

samuel823:
look , in the Docker Container part, i was like what was it

Ice3man543:
What are you facing problem with?

samuel823:
i just dont know what to do after i download subfinder , becuase i didnt see the subfinder there , so that the mean is i cant just run ./subfinder -d (target)

Ice3man543:
What OS are you using? How are you installing it? Provide the exact details!

samuel823:
i got linux , the only problem is i cant run the tools with ./subfinder -d
target.com < becuase there have no subfinder dude

Ice3man543:
Do you know what gopath is? Have you installed golang?

Ice3man543:
Try to read about golang and how to install it and write programs in it. :+1:

samuel823:
yes , but no i didnt

samuel823:
i mean can u just give me the command like after i download golang , what
is the command with run subfinder

samuel823:
ty so much bro

samuel823:
no , i know how to download golang , dude all i need to know is the command
about run the tools

samuel823:
becuase the explain u put in the page is 100% wrong and now im asking

samuel823:
i mean codingo not u , and second maybe u not even better than me , so yea if u dont know anything just shut up pls...

Ice3man543:
@samuel823 The tool repository you're currently viewing, i am the lead author of the tool. It was my idea, i worked on the major parts, created the passive engine, written 16 sources, resolver,bruteforcer, etc all by myself. And now you, who doesn't even knows how to install and run a simple golang tool, who doesn't even knows anything about golang or anything about computers says to me that i don't know more than you.

So what I suggest is that you create your own subfinder. Or any such security tool. And then I'll see what you are.

As far as my explanation is concerned, it's understandable and pretty legible for people who know what they are doing. I didn't intend my tool to be used by **utter noobs** and people who have completely no idea of what they are doing but be arrogant as they know everything. So the person who should better shup the hell up is you. Better control your anger. 

If you can't show respect, get away from here. Do you even know what git is?
I have made a good tool and everyone knows that. And you're not paying me for making this. So, i am not responsible for making sure that every single noob understands how to run this tool.
Thanks
[...]

PR #56, opened 2018-05-28
Beginning foundation of build script

codingo:


Ice3man543:
@codingo You've tested it?

codingo:
Yes, it's currently not building all environments but I'll be making a virtual machine for it soon. Built to scale with us. Given #55 I think having some binaries would be a good thing, but I also think we should be hesitant on making this too easy. We don't want to introduce problems for site owners who have people using this who don't know what they're doing.

Ice3man543:
You're right. Simply running the tool without knowing the consequences is a bad thing. We shouldn't make it really easy to install it. Just people with a bit of understanding should be able to use it.

PR #79, opened 2018-06-22
Added in the suppression of error using silent flag

ProDigySML:
Changed to now have all errors suppressed when using the `-silent` flag. This makes scripting much easier!


Ice3man543:
Thanks for this. Good work :+1:

PR #88, opened 2018-06-25
Banner Update

codingo:
Simple banner update as discussed on slack.


codingo:
Reviewed nny picat (discussed on slack). Merging!

Issue #90, opened 2018-06-25
Docker input list stops


vincentcox:
I am using docker to run subfinder. Individual domains are working, but inputting a list via docker is not working: 

```
$ docker run -it subfinder -dL testdomains.txt -v

[NOTE] Edit /root/.config/subfinder/config.json with your options !
               __    _____           __
   _______  __/ /_  / __(_)___  ____/ /__  _____
  / ___/ / / / __ \/ /_/ / __ \/ __  / _ \/ ___/
 (__  ) /_/ / /_/ / __/ / / / / /_/ /  __/ /
/____/\__,_/_.___/_/ /_/_/ /_/\__,_/\___/_/
                             v0.2 - by @ice3man


Running Source: Ask
Running Source: Archive.is
Running Source: Baidu
Running Source: Bing
Running Source: Censys
Running Source: CertDB
Running Source: CertificateTransparency
Running Source: Certspotter
Running Source: Crt.sh
Running Source: Dnsdb
Running Source: DNSDumpster
Running Source: Dogpile
Running Source: Exalead
Running Source: Findsubdomains
Running Source: Hackertarget
Running Source: Ipv4Info
Running Source: Netcraft
Running Source: PassiveTotal
Running Source: PTRArchive
Running Source: Riddler
Running Source: Securitytrails
Running Source: SSLCertificates
Running Source: Shodan
Running Source: Sitedossier
Running Source: Threatcrowd
Running Source: ThreatMiner
Running Source: Virustotal
Running Source: WaybackArchive
Running Source: Yahoo


```

It just stops there. Any ideas?

Mzack9999:
Thanks for reporting this @vincentcox 

The issue here is that the subfinder executable is run inside the docker container, so it doesn't have access to the testdomains.txt file which resides out of it. In order to run it correctly you have first to mount the local folder into the container.
For example assuming that you are locally in the folder containing the testdomains.txt file:

```bash
$ ls -la
total 8
drwxr-xr-x   3 user  staff   96 25 Jun 14:55 .
drwxr-xr-x  15 user  staff  480 25 Jun 14:54 ..
-rw-r--r--   1 user  staff   18 25 Jun 14:55 testdomains.txt
```

then you can run the dockerized version mounting the local folder into the container in ex. /tmp/subfinder/ in the following way:

```bash
$ docker run  -v `pwd`:/tmp/subfinder -i -t subfinder -dL /tmp/subfinder/testdomains.txt -v

[...]

PR #95, opened 2018-06-26
docker-compose implementation

Mzack9999:
docker-compose skeleton with sample file.
Run with:

docker-compose run subfinder -dL domains.txt -v

This is just a POC, could it be useful? (#90)


codingo:
I think this has now been made redundant, going to close but we can always re-open if we revisit this in future.

PR #102, opened 2018-06-27
Refactored subfinder as embeddable

Mzack9999:


codingo:
Don't forget to bump the version number before merging this. Will add a test case to enforce this soon.

PR #108, opened 2018-07-04
remove lingering reference to SSL certificates in README

picatz:
Dang it, I knew I'd miss something :joy:


picatz:
I also added my twitter link and adjusted @Mzack9999's to be consistent with everyone while I was editing the README. ❤️

PR #116, opened 2018-07-11
Added dep as dependency manager, Created a lighter docker image

anshumanbh:


anshumanbh:
Not sure if you want to include this but I normally run the image like this:

`docker run -v $HOME/.config/subfinder:/root/.config/subfinder -it abhartiya/tools_subfinder -d test.com -t 50 -o /tmp/subfinder.txt -v`

Ice3man543:
@anshumanbh It looks solid work. @Mzack9999 would be able to review this in a couple of days. Then we can merge.

There is one thing - There are a lot of Merge branch master messages in the commit history. It is not a problem, but makes the commit history ugly and clunky. 

Can you clone a fresh version, copy the changes there then make a pr?

anshumanbh:
I'd be happy to once it gets reviewed and approved

Dockerfile
```diff
@@ -1,15 +1,20 @@
...
-# Set an env var that matches your github repo name
-ENV SRC_DIR=/go/src/github.com/Ice3man543/subfinder/
+# Cache the dependencies early
+COPY Gopkg.toml Gopkg.lock ./
+RUN dep ensure -vendor-only -v
 
-# Add the source code:
-ADD . $SRC_DIR
+# Build
+COPY *.go ./
```
Mzack9999: just a little detail, In docker documentation the supported syntax to allow correct layer caching and build seems to be listing files one by one or a directory.
This syntax seems to don't be aware of updates to main.go file, so the intermediate layer never gets updated. Changing it to:

```
COPY main.go ./
``` 
works for me on OSX and Linux.

+1 for using multi-stage builds

anshumanbh:
It does make sense. Thanks for catching that! How would you like to proceed? Should I just clone a fresh repo, make that change and issue a new PR?

Ice3man543:
Yeah I think that's the best thing to do.

anshumanbh:
Okay, closing this for now. Will open a new PR. Thanks!

PR #119, opened 2018-07-13
Fixed typo preventing shodan api key to be set correctly

### LINKED ISSUE #110: subfinder --set-config shodanApiKey=XXXXXXXXXX it doesn't work
The command "subfinder --set-config shodanApiKey=XXXXXXXXXX" it doesn't work probably because is the last item on the config json file. Didn't check the code because I'm almost sure this is the problem. Of course we can add manually but this issue is just to let you know. Thanks a lot for the amazing tool! 👍

Mzack9999:
fixes #110

### DIFF
--- a/subf/subfinder.go
+++ b/subf/subfinder.go
@@ -9,7 +9,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strings"
-
+fv
 	"github.com/subfinder/subfinder/libsubfinder/engines/passive"
 	"github.com/subfinder/subfinder/libsubfinder/helper"
 )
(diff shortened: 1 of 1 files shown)


codingo:
@Mzack9999 Looks like the build for this is failing, could you revisit this?

codingo:
@Mzack9999 I've fixed this as well as resolved a stupid bug I introduced when making the first fix (hence two commits).

@Ice3man543 @picatz if one of you can review this then it should be good to merge.

codingo:
Closes #110

Issue #118, opened 2018-07-13
panic while running job: runtime error: invalid memory address or nil pointer dereference


TvMpt:
The application crashes with some domains, maybe invalid responses are not handled properly somewhere in code...

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: interface conversion: interface {} is runtime.errorString, not string
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x6b9d9b]

goroutine 36 [running]:
github.com/subfinder/subfinder/libsubfinder/helper.(*Pool).subworker.func1(0xc420074af0)
        /home/ec2-user/go/src/github.com/subfinder/subfinder/libsubfinder/helper/pool.go:61 +0x163
panic(0x717da0, 0x93b9f0)
        /usr/local/go/src/runtime/panic.go:502 +0x229
github.com/subfinder/subfinder/libsubfinder/sources/ipv4info.Query(0xc4200e6a60, 0x2, 0x2, 0x68c801, 0x101)
        /home/ec2-user/go/src/github.com/subfinder/subfinder/libsubfinder/sources/ipv4info/ipv4info.go:99 +0x8fb
github.com/subfinder/subfinder/libsubfinder/helper.(*Pool).subworker(0xc420096b40, 0xc420074af0)
        /home/ec2-user/go/src/github.com/subfinder/subfinder/libsubfinder/helper/pool.go:64 +0x6e
github.com/subfinder/subfinder/libsubfinder/helper.(*Pool).worker(0xc420096b40, 0x11)
        /home/ec2-user/go/src/github.com/subfinder/subfinder/libsubfinder/helper/pool.go:78 +0xb3
created by github.com/subfinder/subfinder/libsubfinder/helper.(*Pool).Run
        /home/ec2-user/go/src/github.com/subfinder/subfinder/libsubfinder/helper/pool.go:181 +0x7f
```

codingo:
Can you supply further information? The flags you were using as well as your GO version would greatly aid in diagnosing this.

codingo:
Closing this due to a lack of information. Will re-open this if that changes.

AnotherWayIn:
I too am getting this since the latest update (might not be related, but it's the same on multiple instances). I'm running version go1.10.3

Ice3man543:
@AnotherWayIn could you provide some more context? The domain you were finding, the flags you used, etc.

AnotherWayIn:
[...]

PR #132, opened 2018-08-17
Fix fox DnsDB failure

Mzack9999:
Handles #118


Mzack9999:
Spoke with @picatz on Slack. We are going to merge directly.

PR #178, opened 2019-10-16
Fixing "net/url: invalid control character in URL" error on waybackarchive

icytrues:
Running string() on an integer returns an empty string, changing to strconv.Itoa() instead.

See example: 
```golang
package main

import (
	"fmt"
	"strconv"
)

func main() {
	i := 0
	fmt.Printf("i='%d'\n", i)
	fmt.Printf("i='%s'\n", string(i))
	fmt.Printf("i='%s'\n", strconv.Itoa(i))
}
```
```$ go run test.go
i='0'
i=''
i='0'
```


Ice3man543:
Thanks for the PR @icytrues . Merged it 👍

PR #184, opened 2019-11-18
dnsdumpster: fix csrf

hazcod:


hazcod:
Hmm, seems like I did not have the latest version..

bauthard:
Thanks for the pull request, it's fixed in the latest version.

PR #221, opened 2020-02-05
archiveis - fix panic from closing a closed channel

donners:
If one of the recursive calls encounters an error, it was closing the channel without setting the closed flag. This change moves the channel close outside enumerate() so there's only one place that needs to close it.

Also removed the loop in enumerate() which wasn't necessary.


Ice3man543:
@donners, nice catch. let me review and merge!

PR #229, opened 2020-04-18
Small fix if $HOME/.config does not exists

jonaslejon:
Add -p to mkdir to prevent error message if .config does not exists


bauthard:
Thank you for the pull request @jonaslejon

PR #240, opened 2020-05-18
Added spyse source

akhil-reni:
I was exploring Spyse with a pro user account and found it pretty useful. 

**API documentation:** https://spyse.com/tools/api

**Terms to use:** https://spyse.com/info/terms

### DIFF
--- a/pkg/passive/sources.go
+++ b/pkg/passive/sources.go
@@ -127,7 +127,7 @@ func (a *Agent) addSources(sources []string) {
 		case "sitedossier":
 			a.sources[source] = &sitedossier.Source{}
 		case "spyse":
-			a.sources[source] = &sitedossier.Source{}
+			a.sources[source] = &spyse.Source{}
 		case "sublist3r":
 			a.sources[source] = &sublist3r.Source{}
 		case "threatcrowd":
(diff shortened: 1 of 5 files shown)


pkg/passive/sources.go
```diff
@@ -124,6 +126,8 @@ func (a *Agent) addSources(sources []string) {
 			a.sources[source] = &shodan.Source{}
 		case "sitedossier":
 			a.sources[source] = &sitedossier.Source{}
+		case "spyse":
+			a.sources[source] = &sitedossier.Source{}
```
akhil-reni, line 130: Just noticed - have to change the source name from sitedossier to Spyse.

pkg/subscraping/types.go
```diff
@@ -38,6 +38,7 @@ type Keys struct {
 	PassiveTotalPassword string `json:"passivetotal_password"`
 	Securitytrails       string `json:"securitytrails"`
 	Shodan               string `json:"shodan"`
+	Sypse                string `json:"spyse"`
```
Mzack9999, line 41: This seems a typo, I think it should be `Spyse`

bauthard:
Thank you so much for this pull request @akhil-reni

PR #254, opened 2020-07-01
Update commoncrawl.go

himanshudas:
Updating commoncrawl to include results from year - 2020

### DIFF
--- a/pkg/subscraping/sources/commoncrawl/commoncrawl.go
+++ b/pkg/subscraping/sources/commoncrawl/commoncrawl.go
@@ -22,7 +22,7 @@ type indexResponse struct {
 // Source is the passive scraping agent
 type Source struct{}
 
-var years = [...]string{"2020", "2019", "2018", "2017", "2016"}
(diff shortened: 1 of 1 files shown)


bauthard:
Hi @himanshudas, 

I've checked this PR, and while inspecting results, it doesn't add more numbers but more prone to timeout error because of response time from commoncrawl. 

```
[WRN] Could not run source commoncrawl: context deadline exceeded (Client.Timeout or context cancellation while reading body)
```

do you still believe adding another year will be a good idea?

himanshudas:
Hello @bauthard , 

Thanks a lot for reviewing the commit. I didn't get timeouts during execution. However, commoncrawl as an independent source seems to be slow even without year addition. 

This does add few new subdomains but with a trade-off performance. 

Please, feel feel to close the PR in case you see no value addition 👍

bauthard:
Hi @himanshudas, 

Thank you for the feedback, so I've removed one year from the last to make sure we don't get timed out frequently and keep the latest year to make sure we have the latest results, merging this now.

PR #267, opened 2020-07-15
Handle GitHub Search API rate limit exceeded

### LINKED ISSUE #262: Github Rate Limit
Hi,
as Github is rate limiting requests, I would advise to repeat the new Github recon process around 4-5 times, as suggested by Jason Haddix, with some different sleep time.

Thanks!

vzamanillo:
Fixes 

- #262: Support multiple GitHub tokens to randomly pick one to avoid exceeding request rate limit. Retry after `Retry-After` seconds header value if rate limit abuse is detected. 
- Bug iterating regex matches adds the first match only.

Implements

- Find subdomains using GitHub Search `text-match-metadata` too.
- Distributed use of the GitHub tokens.

### DIFF
--- a/pkg/runner/config.go
+++ b/pkg/runner/config.go
@@ -120,7 +120,7 @@ func (c ConfigFile) GetKeys() subscraping.Keys {
 		keys.DNSDB = c.DNSDB[rand.Intn(len(c.DNSDB))]
 	}
 	if (len(c.GitHub)) > 0 {
-		keys.GitHub = c.GitHub[rand.Intn(len(c.GitHub))]
+		keys.GitHub = c.GitHub
 	}
 
 	if len(c.IntelX) > 0 {
--- a/pkg/subscraping/sources/github/github.go
+++ b/pkg/subscraping/sources/github/github.go
@@ -112,9 +112,11 @@ func (s *Source) enumerate(ctx context.Context, searchURL string, domainRegexp *
 			}
 
 			// Search for domain matches in the code
-			domainMatch := domainRegexp.FindStringSubmatch(string(code))
-			if len(domainMatch) > 0 {
-				results <- subscraping.Result{Source: s.Name(), Type: subscraping.Subdomain, Value: domainMatch[1]}
+			subDomainMatch := domainRegexp.FindAllString(string(code), -1)
+			if len(subDomainMatch) > 0 {
+				for _, subdomain := range unique(subDomainMatch) {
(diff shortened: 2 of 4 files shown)


bauthard:
Hi @vzamanillo, 

Results are surprising, I've found a max number of subdomains for few assets just using this single source, definitely great improvement, few inputs, adding multiple keys from different accounts doesn’t reduce the scan time, so I'm assuming this is expected to avoid any limits and you make use of next keys when the previous one hit the rate limit? 

This is ready to merge after your confirmation.

vzamanillo:
Hi @bauthard!

thank you :).

The strategy is:

- We can add one or more GitHub tokens to the config file.
- The token is always taken randomly to reduce the possibility of a rate limit abuse, it does not matter if you have one only.
- If rate limit abuse is detected we pause the execution for the specified time in the `Retry-After` header received in the GitHub request rate limit response (usually 60 seconds). We do not slow down the requests because I think we have more benefits (speed) waiting 60 seconds while GitHub reset the rate limit count for our token instead forcing a slow down.

Do you think it is correct? Any ideas to improve it?

bauthard:
Hey @vzamanillo, 

I see one place where this can be improved, let me know what you think about it. 

If the user placed more keys in the config file and hit the rate limit, we can pick the next key and continue with the search (if the rate limit is based on the key, not IP) so here we can make use of additional keys from a different account to avoid wait time, if there is no additional keys are stored, follow the current logic of wait time, this might make this process complex maybe, so I'm all good with the current idea for max result.

vzamanillo:
My concern is about how to get the "next" different key I did this because simplifies the process a lot, but you are right, if we have more than one key it would be the best strategy. Let me think about how to manage the keys, I'll do my best.

Ice3man543:
@vzamanillo we could do a simple round robin over all the keys which will distribute the load evenly across all the keys and maintain an internal map of all the keys which have expired and been paused for execution. As soon as we hit github's rate limit, we move the key into the expired bucket. On each run the key is checked if it's in the expired bucket and it's been less than 60 seconds (whatever timeout github has) since it's been there. If yes, delete the key from expired bucket and reuse it again, otherwise move to the next key. Rpeating this process should give us good results.

vzamanillo:
Hey @Ice3man543,

thank you for the proposal and for the help offer! Looks good to me.

bauthard:
Great work on this @vzamanillo, merging this now.

vzamanillo:
Thanks!

PR #269, opened 2020-07-16
Fixed encoded characters in the subdomain results for Common Crawl source

vzamanillo:
We use `QueryUnescape` to unescape the body before finding the subdomains avoding the presence of unwanted characters in the results instead `TrimPrefix` because is case sensitive, we can receive `%2f` or `%2F`.

cw

### DIFF
--- a/pkg/subscraping/sources/commoncrawl/commoncrawl.go
+++ b/pkg/subscraping/sources/commoncrawl/commoncrawl.go
@@ -104,7 +104,7 @@ func (s *Source) getSubdomains(ctx context.Context, searchURL string, domain str
 			}
 			resp.Body.Close()
 
-			src, _ := string(body)
+			src, _ := url.QueryUnescape(string(body))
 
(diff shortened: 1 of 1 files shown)


pkg/subscraping/sources/commoncrawl/commoncrawl.go
```diff
@@ -103,11 +104,10 @@ func (s *Source) getSubdomains(ctx context.Context, url string, domain string, s
 			}
 			resp.Body.Close()
 
-			src := string(body)
+			src, _ := string(body)
 
 			for _, subdomain := range session.Extractor.FindAllString(src, -1) {
 				subdomain = strings.TrimPrefix(subdomain, "25")
```
vzamanillo, line 110: I did not remove this line because I don't know what is for. Does anyone know what char is pretending to remove?

bauthard:
Thank you for spotting and fixing this @vzamanillo

vzamanillo:
Thanks!

PR #271, opened 2020-07-19
Do not run dnsdb source if keys are not specified

### LINKED ISSUE #264: Do not run dnsdb source if keys are not specified
### Current behavior - 

- DNSDB runs even keys are not specified in the config file. 

```bash
$ subfinder -d test.com -sources dnsdb -v

[WRN] Use with caution. You are responsible for your actions
[WRN] Developers assume no liability and are not responsible for any misuse or damage.
[WRN] By using subfinder, you also agree to the terms of the APIs used.

[INF] Enumerating subdomains for test.com
[dnsdb] Source took 1.041913958s for enumeration

[WRN] Could not run source DNSDB: invalid status code received: 403
```

### Expected behavior - 

- Runs all the sources excluding for which keys are not specified in the config file.

vzamanillo:
Fixes #264


bauthard:
thank you @vzamanillo, I can confirm this fixes the issue.

PR #273, opened 2020-07-20
Identify and print which sources need an API key / token when ls is used

vzamanillo:


bauthard:
Looks good @vzamanillo, 

Minor change, we should not display existing keys with `ls`, users can always edit/update the file if needed, reflecting key can leak information sometimes, so do you think we should remove the key's value? 

```
$ ./subfinder -ls

        _     __ _         _         
____  _| |__ / _(_)_ _  __| |___ _ _ 
(_-< || | '_ \  _| | ' \/ _  / -_) '_|
/__/\_,_|_.__/_| |_|_||_\__,_\___|_| v2

		projectdiscovery.io

[WRN] Use with caution. You are responsible for your actions
[WRN] Developers assume no liability and are not responsible for any misuse or damage.
[WRN] By using subfinder, you also agree to the terms of the APIs used.

[INF] Current list of available sources. [28]
[INF] Sources marked with an * needs key or token in order to work.
[INF] You can modify /Users/bauthard/.config/subfinder/config.yaml to configure your keys / tokens.

alienvault
archiveis
binaryedge * 32b68489-xxxx
bufferover
censys
certspotter * 10639_xxxx
certspotterold
commoncrawl
```

vzamanillo:
I thought about it, but I didn't consider it as a leak because on a practical level anyone with permissions on the configuration file could see the keys. In any case, if at first glance it is considered a leak we can eliminate it without problems.

PR #278, opened 2020-07-23
Initial adoption of golangci-lint for continuous integration

vzamanillo:
Implements #277

Needs revision, it should be set as a draft for further improvement before confirming the merge.

There are several linters that I have disabled because for now it is a lot of work to correct the errors, for example

- `funlen`: Detection of long functions.
- `gocyclo`: Calculates  [cyclomatic complexities](https://en.wikipedia.org/wiki/Cyclomatic_complexity).
- `gosec`: Inspects source code for security problems.
- `lll`: Line length linter, used to enforce line length in files.

and I have added some directives to skip the checks with (`nolint`) because they need a small refactor but I think it's fine for now, we can improve it in the future.

Changes need in-depth review to make sure nothing is broken and that they are appropriate.

If you want to test this locally with https://github.com/nektos/act you'll have to add an additional step before `"Set up Go"` due to a dependencies error of the `nektos/act` default docker image.

```
    steps:
      - name: Install dependencies
        run: |
          apt update
          apt install gcc build-essential --assume-yes --quiet --no-install-suggests --no-install-recommends

      - name: Set up Go
        uses: actions/setup-go@v2
        with:
          go-version: 1.13
```
Note I added a new step `golangci-lint` after `Build` step instead of adding a new workflow file as https://github.com/golangci/golangci-lint-action#how-to-use recommends, but in the future we could move it to its own file to have better control over the actions of the GitHub workflow.

I think the configuration is not very strict but it ensures a minimum quality of the code, if you consider it appropriate we can add other linters to make it more strict, but we have to think about it in order to facilitate the work of the collaborators in new pull requests.

We have an open discussion here.

I have do my best, I apologize in advance for any possible problems that may arise.

PR #281, opened 2020-07-29
Sources memory improvement

### LINKED ISSUE #279: Improve sources memory consumption
While doing some memory profiles with `pprof` I've discovered that some sources increase the memory footprint of subfinder in excess ej: `waybackarchive`

wayback1

This is because the size of the results is very large and we are using `ioutil.ReadAll(pagesResp.Body)`.

After some changes to read the response stream using `bufio.NewReader(pagesResp.Body)` the memory consumption is drastically reduced.

wayback2

It happens in other sources too, especially in those that return `json` and no decoder is used to process it, but all the content is put in memory with `ioutil.ReadAll (pagesResp.Body)` and `subdomainExtractor` is used with `regexp` to match subdomains (ej: `threatminer`, `threatcrowd`...).

It would be nice to avoid using `ioutil.ReadAll (pagesResp.Body)` as long as possible and check the rest of the sources to use the `json` responses correctly.

We could do it after merging #278 or we could introduce them directly in that branch.

vzamanillo:
Fixes #279

This branch comes from the same branch of #278 (commit https://github.com/projectdiscovery/subfinder/pull/281/commits/34fb96f0605401144baff0778926d74f178cc13e)

I leave the pull request here to not to lose the local changes, we should merge it after the mentioned pull request we could also put it as a draft to release it after.

Main changes:

Decodes `JSON`  instead of reading all the response content into memory and  parsing with `regex`:

- `bufferover`
- `certspotterold`
- `crtsh`
- `threatcrowd`
- `threatminer`

Read responses using `bufio` instead of `ioutil.ReadAll`

- `github`
- `commoncrawl`
- `hackertarget`
- `waybackarchive`

Fixes triple URL encoding:

- `commoncrawl`
- `waybackarchive`

Check it and tell me what you think!

### DIFF
--- a/pkg/subscraping/sources/alienvault/alienvault.go
+++ b/pkg/subscraping/sources/alienvault/alienvault.go
@@ -22,28 +22,27 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 	results := make(chan subscraping.Result)
 
 	go func() {
+		defer close(results)
+
 		resp, err := session.SimpleGet(ctx, fmt.Sprintf("https://otx.alienvault.com/api/v1/indicators/domain/%s/passive_dns", domain))
 		if err != nil {
 			results <- subscraping.Result{Source: s.Name(), Type: subscraping.Error, Error: err}
 			session.DiscardHTTPResponse(resp)
-			close(results)
 			return
 		}
 
-		otxResp := &alienvaultResponse{}
+		var response alienvaultResponse
 		// Get the response body and decode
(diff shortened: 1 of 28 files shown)


bauthard:
I've removed the **urlscan** source, as they just announced official support of API, so we can add again using official API, Can you rebase this PR with master and confirm if this is ready to merge? @vzamanillo

Ice3man543:
@vzamanillo this PR looks good to me, however before merging could you please squash all the commits into once since their number is quite high. Thanks!

PR #286, opened 2020-08-11
Handle sitedossier 404 when no subdomains are found

vzamanillo:
sitedossier returns a 404 error when no subdomains are found.

### DIFF
--- a/pkg/subscraping/sources/sitedossier/sitedossier.go
+++ b/pkg/subscraping/sources/sitedossier/sitedossier.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math/rand"
+	"net/http"
 	"regexp"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/subscraping/sources/sitedossier/sitedossier.go
```diff
@@ -29,13 +30,19 @@ func (a *agent) enumerate(ctx context.Context, baseURL string) error {
 			return nil
 		default:
 			resp, err := a.session.SimpleGet(ctx, baseURL)
-			if err != nil {
+
+			isnotfound := resp != nil && resp.StatusCode == http.StatusNotFound
```
Ice3man543, line 34: We shuld still read response body even if we get a 404 response since go will not reuse that connection until and unless we read all data from it. Can we change this to allow reading response and then returning?

vzamanillo: Ouch, my mistake, sure.

vzamanillo:
I don't know from where this error comes because that changes are not in my branch... Maybe is due to a missconfiguration of golangci-lint or the workflow.

PR #290, opened 2020-08-11
Added recursive, all flags. More tuning of process

Ice3man543:


pkg/runner/banners.go
```diff
@@ -32,6 +32,20 @@ func (options *Options) normalRunTasks() {
 	if err != nil {
 		gologger.Fatalf("Could not read configuration file %s: %s\n", options.ConfigFile, err)
 	}
+
+	// If we have a different version of subfinder installed
+	// previously, use the new iteration of config file.
+	if configFile.Version != Version {
```
vzamanillo, line 38: This is a breaking change, I've lost my source keys from the config file after running subfinder from source on the master branch. Is this intentional?

Ice3man543: @vzamanillo i could not reproduce the issue. Could you please post steps you took? Since this works on my side.

Ice3man543: All it does is changes the configuration file's source, allsources and recursive-resources files without changing any other keys so i don't think there should be an issue.

pkg/passive/sources.go
```diff
@@ -34,8 +34,48 @@ import (
...
+	"shodan",
+	"spyse",
+	"sublist3r",
+	"threatcrowd",
+	"threatminer",
+	"virustotal",
+}
+
+// DefaultRecursiveSources contains list of default recursive sources
+var DefaultRecursiveSources = []string{
```
vzamanillo, line 62: I think it could be better setting the source as recursive via a property on each source file (e.j: `isRecursive()`) and asking for that property before running instead of using the config file. Any thoughts?

Ice3man543: Yep, i was thinking about that, marking a source as default, recursive etc using functions and doing stuff based on that. Using interfaces would solve that however this was a quick change since the feature is much needed. We could do a bigger refactor doing many more things like `interfaces`, `tests`, etc.  This will be done in the future.

pkg/runner/banners.go
vzamanillo: You are completely right, sorry for this, it was due to some changes in a local testing branch.

Ice3man543: @vzamanillo No problem, it happens all the time :P.

pkg/passive/sources.go
vzamanillo: I am working on a local branch to simplify the key management, I will keep this on mind.

PR #291, opened 2020-08-12
Fixed GitHub search query term + sort results

vzamanillo:
The double quotes arround the search term does not return any results for some terms like "tesla.com".


bauthard:
Thank you for the PR and fix @vzamanillo,  this should cover all the cases.

Issue #301, opened 2020-08-25
[Feature] Request for Sources Addition


Avileox:
#### Sources with API Support:

- [x] alienvault [Api Support]
- [x] zoomeye [Api Support]
- [x] ipv4info [Api Support]
- [x] fofa.so [Api Support]
- [x] intelx.io [Api Support]
- [x] ThreatBook [Api Support]
- [ ] Cloudflare [Api Support]
- [ ] ZETAlytics [Api Support]
- [ ] WhoisXML [Api Support]
- [ ] NetworksDB [Api Support]
- [ ] CIRCL [Api Support]
- [ ] viewdns [Api Support]
- [ ] c99.nl [Api Support]
- [ ] Chinaz [Api Support]

 

#### Sources without API Support:

- [x] ximc
- [x] anubisdb
- [x] robtex
- [x] cebaidu
- [x] riddler.io 
- [ ] ancestry
- [ ] wzpc
- [ ] qianxun
- [ ] ptrarchive
- [ ] ip138
- [ ] mnemonic.no
- [ ] threatminer
- [ ] sonarsearchit.io

#### Some Extra Source for -all enumeration option as they can be time taking:

- [ ] facebook
- [ ] bing [Api Support]
- [ ] googlect [Api Support]
- [ ] baidu
- [ ] pastebin
- [ ] locarchive
- [ ] ukgovarchive
- [ ] yahoo
- [ ] archive.vn

bauthard:
`intelx` is already part of subfinder.

vzamanillo:
ximc is implemented,  WhoisXML not yet.

bauthard:
Thank you @Avileox for the issue and @vzamanillo for adding multiple sources, closing this issue now, as from now on we will be focusing on utilizing existing sources in the best possible way, new sources will be added only when we are getting new results using new sources compared to existing sources.

PR #302, opened 2020-09-02
Implemented fofa source

vzamanillo:
WIP for #301

This is a concept source for `fofa` because I can't test it due to the absence of an API Key.

If anyone has an API Key I invite them to test this branch and would be very grateful.


bauthard:
This needs to verify with a valid key.

bauthard:
@vzamanillo closing this PR, due to the complexity involved to obtain the API key which is required to make use of this source.

PR #304, opened 2020-09-04
added enterprise api support for binaryedge with max pagesize increase

tracertea:
Added support for max pagesize.
Added support for enterprise v1 api subdomains


bauthard:
Tested and Verified.

PR #335, opened 2020-09-27
Implement SonarSearch source

wavvs:
This PR implements Sonar DNS database search from https://sonar.omnisint.io/.


v2/pkg/passive/sources.go
```diff
@@ -59,6 +60,7 @@ var DefaultSources = []string{
 	"riddler",
 	"securitytrails",
 	"shodan",
+	"sonarsearch",
```
vzamanillo, line 63: I think this source should be in the list of recursives, `DefaultRecursiveSources`, the rest is great!

vzamanillo (approved):

bauthard:
Thank you @wavvs for adding this source, @vzamanillo for the review.

PR #341, opened 2020-10-14
Fix CI path

vzamanillo:
`golangci` does not find any issues because the working dir path is wrong and it only analyzes `main.go`


vzamanillo:
Fine, it is working.

bauthard:
Thank you @vzamanillo

PR #342, opened 2020-10-15
Fixed multi-stage build, reduced docker image to 19MB. Was 500MB.

jsav0:
- Remove unnecessary `WORKDIR`
- Correct the multi-stage build that was started originally `AS build-env` 
- Reduced image from 500MB to 19MB


bauthard:
Thank you @jsav0, 

We have also pushed the latest build using this PR to docker hub, that can be pulled using `docker pull projectdiscovery/subfinder:latest`

PR #371, opened 2021-02-15
Fixed shodan module in order to correctly reflect all results

thezoomerr:
At the time of writing this, subdomains are not correctly pulled from Shodan as shown in these pictures:

Direct query to Shodan (`https://api.shodan.io/dns/domain/hackerone.com?key=[API_KEY]`):

image

Results for `subfinder -sources shodan -d hackerone.com`:

image

------------------------------

Looking at the dataset returned by Shodan I noticed that there's a `subdomain` property in the JSON object. Since subfinder doesn't return any additional pieces of information anyway, I believe this is enough to gather the results (instead of looking at the `data` property and extract them from a more complex structure). Here's how the command looks like after changing the code.

Screenshot 2021-02-15 185442

Best, 
@thezoomerr


fnbf (commented):
I tested this one locally and it works great.

ehsandeep (approved):

ehsandeep:
Thank you @thezoomerr for the update, this change definitely make it right and yield complete results from shodan API, changed to the base branch to `dev`.

Thank you @fnbf for the review.

thezoomerr:
No worries. Thank you everyone :)

PR #397, opened 2021-02-22
Improved Github subdomain matching regex

ldionmarcil:
The Regular Expression for matching subdomains did not match on hyphens (`-`) which created a lot of situations where subdomains were being incompletely picked up.

For example, `foo-bar.foobar.com` would only match on `bar.foobar.com`. Also added a match on only-alphanum for the first character, as subdomains can't start with `-`.


Mzack9999 (approved):

ehsandeep:
Thank you @ldionmarcil for improving the regex.

PR #403, opened 2021-03-21
fix(crtsh): typo preventing results being returned fixed

infosec-au:
There's an extra % in the crt.sh API endpoint that prevents results from being returned. I've fixed this typo so that this source works again.

### DIFF
--- a/v2/pkg/subscraping/sources/crtsh/crtsh.go
+++ b/v2/pkg/subscraping/sources/crtsh/crtsh.go
@@ -72,7 +72,7 @@ func (s *Source) getSubdomainsFromSQL(domain string, results chan subscraping.Re
 }
 
 func (s *Source) getSubdomainsFromHTTP(ctx context.Context, domain string, session *subscraping.Session, results chan subscraping.Result) bool {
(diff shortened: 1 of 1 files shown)


ehsandeep:
@infosec-au thank you for the PR, can you share an instance where you experienced crt.sh API endpoint that prevents results from being returned? can't repo this side.

```bash
subfinder -version
[INF] Current Version: 2.4.6
```

```bash
subfinder -d hackerone.com -sources crtsh -silent

support.hackerone.com
docs.hackerone.com
hackerone.com
links.hackerone.com
mta-sts.forwarding.hackerone.com
mta-sts.managed.hackerone.com
www.hackerone.com
api.hackerone.com
go.hackerone.com
info.hackerone.com
mta-sts.hackerone.com
```

ehsandeep:
@infosec-au thank you for opening a PR for this, we identified the root cause of this behavior, as the default subfinder crtsh source supports DB connection to query, if failed it fallback to web endpoint which was not working due to a bug we fixed in following PR https://github.com/projectdiscovery/subfinder/pull/410, now it should work with the latest release of subfinder.

Closing this PR as the fix implemented in referenced PR.

PR #416, opened 2021-04-17
Add example of how to use subfinder as a Go library

hachh:
@ehsandeep I added a basic example of how to use subfinder as a Go library


ehsandeep:
thank you @hachh

PR #442, opened 2021-08-09
Add http-proxy flag and C99 to README.md

taythebot:
I've added the `http-proxy` flag and C99 source to the readme file. I forgot to add them in when I added the features.


ehsandeep (approved):

ehsandeep:
thank you @taythebot

PR #445, opened 2021-08-09
Retrieve archiveis first page

maldevel:
Archiveis is not working properly. When you are trying to retrieve the next page (e.g. https://archive.is/offset=50/*.github.com) you are getting an HTTP error 429 in golang because it asks for the user's input (recaptcha challenge solving). However, instead of removing the website completely you could just retrieve the results of the first page only.

### DIFF
--- a/v2/pkg/subscraping/sources/archiveis/archiveis.go
+++ b/v2/pkg/subscraping/sources/archiveis/archiveis.go
@@ -9,12 +9,15 @@ import (
 	"github.com/projectdiscovery/subfinder/v2/pkg/subscraping"
 )
 
+// Source is the passive scraping agent
 type Source struct{}
 
+// Name returns the name of the source
 func (s *Source) Name() string {
 	return "archiveis"
 }
 
+// Run function returns all subdomains found with the service
 func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Session) <-chan subscraping.Result {
 	results := make(chan subscraping.Result)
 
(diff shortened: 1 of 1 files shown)


ehsandeep (commented):
@maldevel I see `503` status instead of `429`, is it expected after your change?

```sh
git branch --show-current
pr/445


./subfinder -d hackerone.com -v -sources archiveis

               __    _____           __         
   _______  __/ /_  / __(_)___  ____/ /__  _____
  / ___/ / / / __ \/ /_/ / __ \/ __  / _ \/ ___/
 (__  ) /_/ / /_/ / __/ / / / / /_/ /  __/ /    
/____/\__,_/_.___/_/ /_/_/ /_/\__,_/\___/_/ v2.4.8

		projectdiscovery.io

Use with caution. You are responsible for your actions
Developers assume no liability and are not responsible for any misuse or damage.
By using subfinder, you also agree to the terms of the APIs used.

[INF] Enumerating subdomains for hackerone.com
[WRN] Could not run source archiveis: unexpected status code 503 received from https://archive.is/*.hackerone.com
```

maldevel:
I just tested it and it works. The 503 error means that the service wasn't available. Maybe it was temporarily unavailable on the server side? The HTTP 429 error code is being received when you try to capture the next page and it happens because of the recaptcha challenge.

maldevel:
Sometimes you might receive http 429 error or other http errors because of the user agent string. The websites (not only archiveis) are blacklisting certain user agent strings and/or identical get/post requests that are being repeated in a short period of time. In that case, to solve this issue you have to change the user agent string in the agent.go file or preferably use a random user agent string with each http request.

Mzack9999 (approved):

ehsandeep:
thank you @maldevel, I've created a separate issue to support user agent with each HTTP request.

PR #446, opened 2021-08-10
v2.4.9 Preparation

ehsandeep:


maldevel:
I just tested it and it works. The 503 error means that the service wasn't available. Maybe it was temporarily unavailable on the server side? The HTTP 429 error code is being received when you try to capture the next page because of the recaptcha challenge.

Issue #447, opened 2021-08-10
Use a random user agent string with each http request.


ehsandeep:
Sometimes you might receive http 429 error or other http errors because of the user agent string. The websites (not only archiveis) are blacklisting certain user agent strings and/or identical get/post requests that are being repeated in a short period of time. In that case, to solve this issue you have to change the user agent string in the agent.go file or preferably use a random user agent string with each http request.

_Originally posted by @maldevel in https://github.com/projectdiscovery/subfinder/issues/445#issuecomment-895982601_

ehsandeep:
We already have this support in nuclei and httpx project.

taythebot:
I'm working on implementing this feature and was wondering if we should keep the default user agent `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36` or change it to `subfinder - Open-source Project (github.com/projectdiscovery/subfinder)` to match httpx?

taythebot:
I've gone ahead and added the `unsafe` flag which uses the default user agent. If the `random-agent` flag is required please let me know!

Issue #362, opened 2021-01-12
[Feature] rate limit support for http requests


ZeroDot1:
Add a function to make a random delay between the requests.
The user should have the possibility to set a time delay, e.g. a random delay from 1 to 10 seconds.
This function helps not to be blocked directly by the different services when you make a lot of requests.

ehsandeep:
For this, a new flag, say `delay`, can be added to ensure that the user-defined delay between each HTTP request sent from SubFinder is respected.

```
-delay         Seconds to wait between requests (default 0)
```

Example use:-

```
subfinder -d example.com -delay 2
```

PR #449, opened 2021-08-11
Add Rate Limit and User Agent randomization for HTTP requests

taythebot:
Added user agent randomization(#447) and rate limiting for HTTP requests (#362).

User agent randomization is implemented using the [uarand ](github.com/corpix/uarand) library. It is enabled by default since Subfinder exclusively uses third party APIs.

Rate limit is implemented using the [ratelimit](https://github.com/uber-go/ratelimit) library and the flag is `rate-limit`. The default value for the flag is 0, which creates an "unlimited" rate limit instance.


ehsandeep (approved):

ehsandeep:
thank you @taythebot for adding both features.

PR #450, opened 2021-08-11
Modify Dockerfile

EONRaider:
This PR modifies the Dockerfile in the following ways:
- Add deterministic image names
- Add a non-privileged user to the container. The importance of using a least privileged user is well-known, but you can check the advice from [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html) and [Snyk](https://snyk.io/blog/10-docker-image-security-best-practices/) if necessary. At the least, enabling such user makes it easy for those that use the Docker image and need to publish volumes (files are not written by root, making things easier)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,20 @@
-# Build
-FROM golang:1.16.7-alpine AS build-env
+# Build image: golang:1.16.6-alpine3.14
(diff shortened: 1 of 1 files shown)


ehsandeep:
@EONRaider closing the PR for the same reason mentioned here - https://github.com/projectdiscovery/subfinder/pull/432

PR #464, opened 2021-10-07
Update README.md

mdhama:
Getting error while installing 

`package github.com/projectdiscovery/subfinder/v2/cmd/subfinder@latest: cannot use path@version`


ehsandeep:
Closing as per above comment.

PR #476, opened 2021-12-17
Source full hunt.io

LuitelSamikshya:


v2/pkg/subscraping/sources/fullhunt/fullhunt.go
```diff
@@ -0,0 +1,59 @@
...
+// Source is the passive scraping agent
+type Source struct{}
+
+func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Session) <-chan subscraping.Result {
+	results := make(chan subscraping.Result)
+
+	go func() {
+		defer close(results)
+		
+		resp, err := session.Get(ctx, fmt.Sprintf("https://fullhunt.io/api/v1/domain/%s/details", domain), "", map[string]string{"X-API-KEY": session.Keys.FullHunt})
```
Ice3man543, line 32: Instead of the details endpoint, we should use [subdomains](https://api-docs.fullhunt.io/#get-subdomains-of-a-domain) api.

LuitelSamikshya: Done.

ehsandeep (approved):

PR #477, opened 2021-12-20
Add source zoomeye subdomain api 

M09Ic:
Use new api(https://api.zoomeye.org/domain/search) instead of zoomeye search.

It can find more subdomains than zoomeye search.this api is free, fast and no limit

Need to get apikey on this page: https://www.zoomeye.org/profile/info

Set "zoomeyekey" in config.yaml:

```
zoomeyeapi:
    - c659bb1e-0000-00000-0000-c9643450fee7
```

example:
subfinder -d shein.com -sources [source]

zoomeyeapi
`Found 74 subdomains for shein.com in 3 seconds 211 milliseconds`

zoomeye
`Found 40 subdomains for shein.com in 47 seconds 396 milliseconds`

### DIFF
--- a/v2/pkg/subscraping/sources/zoomeyeapi/zoomeyeapi.go
+++ b/v2/pkg/subscraping/sources/zoomeyeapi/zoomeyeapi.go
@@ -59,15 +59,13 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 			err = json.NewDecoder(resp.Body).Decode(&res)
 			if err != nil {
 				results <- subscraping.Result{Source: s.Name(), Type: subscraping.Error, Error: err}
-				resp.Body.Close()
+				_ = resp.Body.Close()
 				return
 			}
-			resp.Body.Close()
 			pages = int(res.Total/1000) + 1
 			for _, r := range res.List {
 				results <- subscraping.Result{Source: s.Name(), Type: subscraping.Subdomain, Value: r.Name}
 			}
-			currentPage++
 		}
 	}()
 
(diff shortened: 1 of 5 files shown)


M09Ic:
additional:
this commit https://github.com/projectdiscovery/subfinder/pull/477/commits/341575cce20c9a41a8ca7ffd6c834c7306d55794

fix zoomeye source(not zoomeyeapi) turn page bug.


test:
-d shein.com -sources zoomeye

Before repair:
Found 40 subdomains for shein.com in 47 seconds 396 milliseconds

After repair:
Found 55 subdomains for shein.com in 1 minute 30 seconds

ehsandeep (commented):
@M09Ic thank you for adding API support, as the API produces better results, do you think we can consider deprecating `zoomeye` source and replacing the API with it, or do you have any additional value by keeping the old source?

M09Ic:
> @M09Ic thank you for adding API support, as the API produces better results, do you think we can consider deprecating `zoomeye` source and replacing the API with it, or do you have any additional value by keeping the old source?

zoomeye and zoomeyeapi have different data sources.

zoomeye fetches subdomains from ip, more like fofa and shodan, but there is a limit of 400 data items for normal user.

zoomeyeapi is querying subdomain data directly from the api, rather than finding subdomains from the ip. And zoomeyeapi has no limit.

So, zoomeye has a different data source than zoomeyeapi, and may have some additional subdomians. I think we can keep the zoomeye source

ehsandeep (approved):

v2/pkg/subscraping/sources/zoomeyeapi/zoomeyeapi.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/projectdiscovery/subfinder/v2/pkg/subscraping"
+)
+
+//type loginResp struct {
```
Mzack9999, line 12: Would it be possible to remove the commented code if not necessary?

Mzack9999 (changes requested):
Looks good! There is only some commented code to remove if possible.

v2/pkg/subscraping/sources/zoomeyeapi/zoomeyeapi.go
M09Ic: yes. I copy it from zoomeye and forget to remove some comment.

PR #484, opened 2022-01-17
get intimations when the process is done.

AbhijithAJ:
When I run this script, I frequently check if it is complete. 
So I have developed a [intimations](https://github.com/AbhijithAJ/intimations) module which will intimate me via a telegram message once it is done.

Hope this will help others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,13 @@ ip addr
 subfinder -d hackerone.com -b 192.168.1.87
 ```
 
+##### TIP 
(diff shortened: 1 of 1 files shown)


ehsandeep:
Hi @AbhijithAJ, thanks for letting me know about this project, additonally we have also published [notify](https://github.com/projectdiscovery/notify) including support for multiple providers, with that said it's not something we wanted to include in each proejct readme.

PR #505, opened 2022-02-23
fix(fofa): update url in source fofa

zt2:
fofa.so is deprecated, now is fofa.info.


ehsandeep:
thank you @zt2, this should be already updated with https://github.com/projectdiscovery/subfinder/pull/496/

Issue #519, opened 2022-03-17
Add source domain to json output


zerodivisi0n:
### Please describe your feature request:
Add `input` field with source domain to the json output.
E.g. current output:
```
$ echo hackerone.com | subfinder -silent -json -collect-sources
{"host":"mta-sts.managed.hackerone.com","sources":["crtsh","anubis","alienvault"]}
{"host":"resources.hackerone.com","sources":["alienvault"]}
...
```
Output with feature:
```
echo hackerone.com | subfinder -silent -json -collect-sources
{"host":"mta-sts.managed.hackerone.com","input":"hackerone.com","sources":["crtsh","anubis","alienvault"]}
{"host":"resources.hackerone.com","input":"hackerone.com","sources":["alienvault"]}
...
```
By the way, Amass returns such a field in the output. Why not do the same in Subfinder.

If you don't mind, I'm ready to implement this feature.

### Describe the use case of this feature:
When enumerating multiple domains it is impossible match output domains with the input. This can be useful when enumerating a large number of unrelated domains.

ehsandeep:
@zerodivisi0n thanks for suggesting this useful feature, and please feel free to open a PR for this when possible.

zerodivisi0n:
@ehsandeep Sorry, I missed your reaction. Just opened PR #523 .

PR #523, opened 2022-03-24
Add input field to json output

zerodivisi0n:
This PR implements feature request #519
- In the first commit, I moved all JSON types into one place - removed JSON tags from `resolve.HostEntry` and add a new type `jsonSourceResult`. This is necessary to keep all the JSON output in one place.
- In the second commit I added `Input` type to all `json*` result types. As a simple solution, I decided to pass the `input` domain as a separate argument to all outputter functions.

### DIFF
--- a/v2/pkg/resolve/resolve.go
+++ b/v2/pkg/resolve/resolve.go
@@ -25,8 +25,8 @@ type ResolutionPool struct {
 
 // HostEntry defines a host with the source
 type HostEntry struct {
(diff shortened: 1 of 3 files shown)


ehsandeep (approved):

Mzack9999 (approved):

ehsandeep:
thank you @zerodivisi0n for adding this support 🙌🏻

PR #527, opened 2022-03-30
Don't make missing providers-config fatal on first-run

thesamesam:
It's okay if we fall back to the default config
which is in memory & will get saved out when done
parsing.

This is on top of the work in #526, but a slightly
different issue.

See: #526
Signed-off-by: Sam James <sam@gentoo.org>


Mzack9999 (approved):

thesamesam:
Thanks!

PR #582, opened 2022-07-25
Match/filter

### LINKED ISSUE #575: Filter the output based on a given scope
There might be cases where the scope of a bug bounty program is limited to a certain set of subdomain. By adding a new flag to filter the results of subfinder based on this scope, we can reduce the amount of traffic to be passed into the next tools and only focus on in-scope items.

CLI Flag:

```yaml
   -m, -match       match output based on the specified pattern
   -f, -filter      filter output based on the specified pattern
```

Example run:

```
subfinder -d example.com -match "*.api,*waf,login-*" -silent
subfinder -d example.com -match scope.txt -silent
```

Example `scope.txt` file (both exact matches and wildcards should be supported):
```
*.static.example.com
reg.example.com
*.static
*.api-*
reg
```

```sh
subfinder -d example.com -match scope.txt -silent

reg.example.com
69.48.148.215.static.example.com
69.48.134.63.static.example.com
69.48.129.64.static.example.com
69.48.129.82.static.example.com
69.48.134.12.static.example.com
69.48.142.18.static.example.com
69.48.134.23.static.example.com
69.48.134.137.static.example.com
69.48.134.201.static.example.com
69.48.129.155.static.example.com
69.48.129.130.static.example.com
69.48.142.22.static.example.com
```

LuitelSamikshya:
closes #575


v2/pkg/runner/enumerate.go
```diff
@@ -147,3 +147,36 @@ func (r *Runner) EnumerateSingleDomain(ctx context.Context, domain string, outpu
 
 	return nil
 }
+func (r *Runner) findMatchingSubdomain(pattern, value string) (bool, error) {
+	pattern = strings.ReplaceAll(pattern, ".", "\\.")
```
Mzack9999, line 151: Precompile and validate match/filtering regexes/keywords in [validateOptions](https://github.com/projectdiscovery/subfinder/blob/cb6fbbb59f1af05c1b50e2b05320d7540ac3aadc/v2/pkg/runner/validate.go#L12), they can be stored in a private slice field of `Options`, for example:
```go
type Options struct {
...
matchRegexes      []*regexp.Regexp
filterRegexes []*regexp.Regexp
...
}
```

v2/pkg/runner/validate.go
```diff
@@ -34,6 +34,10 @@ func (options *Options) validateOptions() error {
 		return errors.New("hostip flag must be used with RemoveWildcard option")
 	}
 
+	// Match or Filter are not allowed together
```
Mzack9999, line 37: Similarly to `httpx`, I think the tool should support both matching and filtering, sequentially [applying them](https://github.com/projectdiscovery/httpx/blob/e4552498194fd46eca2b8984aad9cc6d49b3cb58/runner/runner.go#L575)

Mzack9999 (changes requested):

v2/pkg/runner/enumerate.go
LuitelSamikshya: Done

v2/pkg/runner/validate.go
LuitelSamikshya: Done

Mzack9999 (approved):

Mzack9999:
```console
$ go run . -d example.com -match "megalom.*" -silent
megalom.example.com
$ go run . -d example.com -filter "megalom.*" -silent | wc -l
   22800
```

forgedhallpass (changes requested):
This functionality requires extensive tests with positive and negative scenarios!

v2/pkg/runner/enumerate_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"testing"
+)
+
+func TestFilterAndMatchSubdomainForMatch(t *testing.T) {
+	options := &Options{}
+	options.Domain = []string{"example.com"}
+	options.Threads = 10
+	options.Timeout = 10
+	options.Output = os.Stdout
+	options.Match = []string{"*.example.com"}
```
Mzack9999, line 15: I think we should cover a few more cases in detail:
- Literals (e.g. `reg.example.com`)
- Multiple wildcards (e.g. `*.api-*`)
- Sequential Cascade Matchers
- Sequential Cascade Filters
- Matchers together with filters

Mzack9999 (changes requested):

v2/pkg/runner/enumerate_test.go
LuitelSamikshya: Done

ehsandeep (approved):

Mzack9999 (changes requested):
The following test will fail:
```go
t.Run("Filter and Match", func(t *testing.T) {
	options.Filter = []string{"example.com"}
	options.Match = []string{"www.example.com"}
	err := options.validateOptions()
	if err != nil {
		t.Fatalf("Expected nil got %v while validation\n", err)
	}
	runner, err := NewRunner(options)
	if err != nil {
		t.Fatalf("Expected nil got %v while creating runner\n", err)
	}
	subdomain := []string{"example.com", "www.example.com"}
	for _, sub := range subdomain {
		match := runner.filterAndMatchSubdomain(sub)
		require.True(t, match, "Expecting a boolean True value ")
	}
})
```
While compiling the regexes, I think we need to use anchors (`^` and `$`) for those not containing wildcards (e.g. `example.com` => `^example\.com$`). Otherwise, all strings containing the regular expression value will match.

Mzack9999 (approved):

ehsandeep (approved):

PR #588, opened 2022-08-10
Sources refactor and bug fixes

forgedhallpass:


ehsandeep (commented):
@forgedhallpass 

```
echo example.com | ./subfinder -s DNSDB -v
echo example.com | ./subfinder -s dnsdb -v
```

Both results into

```
[WRN] There is no source with the name: 'dnsdb'
```

forgedhallpass:
> @forgedhallpass
> 
> ```
> echo example.com | ./subfinder -s DNSDB -v
> echo example.com | ./subfinder -s dnsdb -v
> ```
> 
> Both results into
> 
> ```
> [WRN] There is no source with the name: 'dnsdb'
> ```

Good catch. That was the only source that was not lower case. The issue is now fixed and I've also added extra debug information to help with the troubleshooting.

Case 1 - Missing API key for the selected source:

image

Case 2 - Bad API key:

image

ehsandeep (approved):

v2/pkg/runner/options.go
```diff
@@ -203,27 +193,28 @@ func ParseOptions() *Options {
...
+	// at the end of options parsing from default via goflags.
+	if err := UnmarshalFrom(location); isFatalErr(err) && !errors.Is(err, os.ErrNotExist) {
+		gologger.Fatal().Msgf("Could not read providers from '%s': %s\n", location, err)
 	}
+}
 
-	// unmarshal fields to migrate into temporary struct
-	if err := yaml.NewDecoder(fileSource).Decode(providers); isFatalErr(err) {
+func migrateToProviderConfig(source, dest string) error {
+	fileSource, err := os.Open(source)
```
Mzack9999, line 211: Previously `source` was unmarshaled into a temporary providers struct, then marshaled to the destination. Here it seems it's not used anymore?

Mzack9999 (changes requested):

v2/pkg/passive/sources.go
```diff
@@ -39,185 +45,103 @@ import (
...
-		delete(a.sources, source)
+	if useSourcesSupportingRecurse {
+		for sourceName, source := range sources {
+			if !source.HasRecursiveSupport() {
+				delete(sources, sourceName)
+			}
+		}
 	}
+
+	gologger.Debug().MsgFunc(func() string {
```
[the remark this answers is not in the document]
forgedhallpass: Why? The whole point of `MsgFunc` is to only evaluate code, if the verbose or debug flag was explicitly requested.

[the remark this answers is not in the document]
Mzack9999: Asynchronous access to referenced map accessed. Inline sequential eventuation will avoid potential future data races.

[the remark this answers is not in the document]
forgedhallpass: I might be missing something. Where do you see asynchronous access?

v2/pkg/runner/options.go
forgedhallpass: image

v2/pkg/subscraping/utils.go
```diff
@@ -19,12 +25,34 @@ func NewSubdomainExtractor(domain string) (*regexp.Regexp, error) {
...
-		if v == key {
-			return true
+func PickRandom[T any](v []T, sourceName string) T {
+	var result T
+	length := len(v)
+	if length == 0 {
+		gologger.Debug().Msgf("Cannot use the '%s' source because there was no API key/secret defined for it.", sourceName)
+		return result
+	}
+	return v[rand.Intn(length)]
```
Ice3man543, line 35: we should seed the random number generator first or the items picked won't be random by `rand.Intn`

v2/pkg/subscraping/types.go
```diff
@@ -20,54 +20,35 @@ type Source interface {
...
+
+	// HasRecursiveSupport returns true if the current source
+	// accepts subdomains (e.g. subdomain.domain.tld),
+	// not just root domains.
+	HasRecursiveSupport() bool
+
+	// NeedsKey returns true if the source requires an API key
+	NeedsKey() bool
+
+	AddApiKeys([]string)
```
Ice3man543, line 38: couldn't we have the API keys stored somewhere instead of having a method that stores these values per provider? it just seems more elegant to have them stored somewhere as we had previously and picked randomly without this being done at time of initialization

v2/pkg/subscraping/types.go
```diff
@@ -20,54 +20,35 @@ type Source interface {
...
 }
 
 // Session is the option passed to the source, an option is created
 // uniquely for each source.
 type Session struct {
 	// Extractor is the regex for subdomains created for each domain
 	Extractor *regexp.Regexp
-	// Keys is the API keys for the application
-	Keys *Keys
```
Ice3man543, line 33: this approach of storing and retreiving keys actually is simpler and easier to maintain compared to the generics one we have implemented now. Any specific reason for the current implementation? This seems to overcomplicate things

v2/pkg/subscraping/types.go
```diff
@@ -20,54 +20,35 @@ type Source interface {
...
 	// Client is the current http client
 	Client *http.Client
 	// Rate limit instance
 	RateLimiter ratelimit.Limiter
 }
 
-// Keys contains the current API Keys we have in store
```
Ice3man543, line 40: same here

v2/pkg/subscraping/utils.go
forgedhallpass: I haven't touched this part directly. The previous code used `sliceutil.PickRandom` which does exactly this. I've just extended it and made it generic.

```go
func PickRandom(v []string) string {
	return v[rand.Intn(len(v))]
}
```

v2/pkg/subscraping/types.go
forgedhallpass: No, that's the whole point. Sources should be self-contained. About elegance, I believe my approach is far more elegant than the previous one, not to mention that it's significantly less error prone and easier to add new sources.

v2/pkg/subscraping/types.go
forgedhallpass: I strongly disagree. See my comment above :)

v2/pkg/subscraping/utils.go
Ice3man543: sliceutil does call rand.Seed internally automatically iirc

v2/pkg/subscraping/types.go
Ice3man543: i think there a few major problems with this approach. I like the AddApiKeys function but the way we implement the addition is an issue, mainly because the api keys are stored as global variables. which means if we want to use multiple  instances of source we cannot do it. Instead it should be stored as a variable in source itself and probably that'll be better. Also i see the function code related to API keys has been copy pasted a lot so maybe that is something we should fix as well

v2/pkg/passive/sources.go
```diff
@@ -39,185 +46,102 @@ import (
...
-	"certspotter",
-	"crtsh",
-	"dnsdumpster",
-	"hackertarget",
-	"passivetotal",
-	"securitytrails",
-	"sonarsearch",
-	"virustotal",
-}
+var NameSourceMap = make(map[string]subscraping.Source, len(AllSources))
```
Ice3man543, line 88: the make statement should be a part of the init function instead of being called globally

v2/pkg/subscraping/sources/github/github.go
```diff
@@ -38,18 +39,21 @@ type response struct {
 // Source is the passive scraping agent
 type Source struct{}
 
+var apiKeys []string
```
Ice3man543, line 42: this should be stored within source variable instead of as a global. and this is relevant for all sources

v2/pkg/subscraping/utils.go
forgedhallpass: > `sliceutil` does call `rand.Seed` internally automatically iirc

Actually it doesn't, but I'll update my code, because it's a good point.

v2/pkg/subscraping/types.go
forgedhallpass: The keys are now moved inside the `Source` struct as discussed. The code related to API keys is not copy-pasted, but uses a generic method that enables naming the key values individually for each source.

v2/pkg/subscraping/sources/github/github.go
forgedhallpass: done

v2/pkg/subscraping/utils.go
```diff
@@ -32,6 +33,7 @@ func PickRandom[T any](v []T, sourceName string) T {
 		gologger.Debug().Msgf("Cannot use the '%s' source because there was no API key/secret defined for it.", sourceName)
 		return result
 	}
+	rand.Seed(time.Now().UnixNano())
```
Ice3man543: this does not need to be in this function. you can just have an init() function and call it there once

Ice3man543: calling it each time is redundant and unncessary

Ice3man543 (approved):

Issue #649, opened 2022-09-26
Create/aggregate tests for providers that do NOT require an API key


forgedhallpass:
It seems that we can test all the providers that do not require an API key, with one simple logic that iterates through all and does some assertions like in https://github.com/projectdiscovery/subfinder/pull/648.

dogancanbakir:
I can work on this.

forgedhallpass:
> I can work on this.

Thank you, I've assigned it to you.

PR #651, opened 2022-09-26
Integration test for sources w/o auth #649

dogancanbakir:
This PR adds integration tests for sources w/o [auth](https://github.com/projectdiscovery/subfinder/issues/649)


Ice3man543 (changes requested):
@dogancanbakir the tests are failing for some reason. can you please fix them?

forgedhallpass:
I've refactored the test to make detecting problems easier (will commit it later). The point is that based on that I've noticed that there are 3 or 4 providers that are not working currenlty. One has anti-automation (captcha) and 2 or 3 other services are not accessible anymore (gateway error and temporarily unavailabe).

dogancanbakir:
@Ice3man543  As @forgedhallpass  pointed out, some of the providers are not working. Additionally, we should consider separating integration tests from the current pipeline, and maybe running them as a cron job might be a good idea.

Ice3man543:
@forgedhallpass we should create issues regarding failing providers and ideally remove them since querying them is a waste of time.

forgedhallpass:
> @forgedhallpass we should create issues regarding failing providers and ideally remove them since querying them is a waste of time.

That's was the plan, but only for tomorrow :)

PR #652, opened 2022-09-28
Create/aggregate tests for providers that do NOT require an API key 

forgedhallpass:


forgedhallpass:
For the tests to pass, it requires completion of #654, in which non-functional sources has been removed.

ehsandeep (changes requested):
Merge conflict

forgedhallpass:
The test now fails because `commoncrawl` is not stable. Sometimes it work, sometimes it doesn't...

```
<html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
<hr><center>nginx/1.18.0 (Ubuntu)</center>
</body>
</html>
<!-- a padding to disable MSIE and Chrome friendly error page -->
```

Issue #671, opened 2022-10-17
Segfault on '[' input


r0nk:
**Describe the bug**
Subfinder crashes when [ is passed on standard input. This is due to how regex compilation errors are handled.

**Subfinder version**
v2.5.4

**Complete command you used to reproduce this**
`echo [ | subfinder`

[standard error output](https://github.com/projectdiscovery/subfinder/files/9803606/log.txt)

PR #672, opened 2022-10-17
Prevent EnumerateSubdomains() from continuing if there is an error in Regex

r0nk:


r0nk:
Should fix https://github.com/projectdiscovery/subfinder/issues/671#issue-1412049848

v2/pkg/passive/passive.go
```diff
@@ -17,6 +17,8 @@ func (a *Agent) EnumerateSubdomains(domain string, proxy string, rateLimit, time
 		session, err := subscraping.NewSession(domain, proxy, rateLimit, timeout)
 		if err != nil {
 			results <- subscraping.Result{Type: subscraping.Error, Error: fmt.Errorf("could not init passive session for %s: %s", domain, err)}
+			close(results)
```
Mzack9999, line 20: I guess we can put a `defer close(results)` at the beginning of the go routine

v2/pkg/passive/passive.go
```diff
@@ -52,6 +54,5 @@ func (a *Agent) EnumerateSubdomains(domain string, proxy string, rateLimit, time
 		close(results)
```
Mzack9999, line 54: redundant if previous suggestion is implemented

Mzack9999 (changes requested):

Mzack9999 (approved):

Ice3man543 (approved):

PR #699, opened 2022-11-09
Add callback function

ShubhamRasal:
```
package main

import (
	"bytes"
	"fmt"
	"io"
	"log"

	"github.com/projectdiscovery/subfinder/v2/pkg/resolve"
	"github.com/projectdiscovery/subfinder/v2/pkg/runner"
)

func main() {
	mycallback := func(s *resolve.HostEntry) {
		fmt.Println("Processing:", s.Host, " Source:", s.Source)
	}
	runnerInstance, _ := runner.NewRunner(&runner.Options{
		Threads:            10,                       // Thread controls the number of threads to use for active enumerations
		Timeout:            30,                       // Timeout is the seconds to wait for sources to respond
		MaxEnumerationTime: 10,                       // MaxEnumerationTime is the maximum amount of time in mins to wait for enumeration
		Resolvers:          resolve.DefaultResolvers, // Use the default list of resolvers by marshaling it to the config
		OnResult:           mycallback,
	})

	buf := bytes.Buffer{}
	err := runnerInstance.EnumerateSingleDomain("projectdiscovery.io", []io.Writer{&buf})
	if err != nil {
		log.Fatal(err)
	}

	data, err := io.ReadAll(&buf)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("%s", data)
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ func main() {
 		Timeout:            30,                       // Timeout is the seconds to wait for sources to respond
 		MaxEnumerationTime: 10,                       // MaxEnumerationTime is the maximum amount of time in mins to wait for enumeration
 		Resolvers:          resolve.DefaultResolvers, // Use the default list of resolvers by marshaling it to the config
-		ResultCallback: func(s *resolve.HostEntry) {
+		ResultCallback: func(s *resolve.HostEntry) {  // Callback function to execute for available host
 			log.Println(s.Host, s.Source)
 		},
 	})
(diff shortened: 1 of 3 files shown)


Mzack9999 (changes requested):
lgtm - Missing things:
- [ ] Update docs (use as a library section)
- [x] Check if any integration test covering library usage needs to be updated

Mzack9999 (approved):

ehsandeep (approved):

PR #701, opened 2022-11-09
Remove sonar source

vzamanillo:
omnisint.io from Sonar Labs is dead, is not working anymore, this is causing test build failures.

image


vzamanillo:
I see this was done in the past #657 ^^

ehsandeep (approved):

PR #702, opened 2022-11-09
Fixed source tests, added crtsh with recursion

vzamanillo:
I didn't have time to correct it before the merge


ehsandeep (approved):

vzamanillo:
Noticed the threatminer API for subdomains check is not working right now

image


image

PR #756, opened 2023-02-09
Add ctx param in agent enumerate method

EndPositive:
so that I can cancel the enumeration manually too

### DIFF
--- a/v2/pkg/passive/passive.go
+++ b/v2/pkg/passive/passive.go
@@ -10,8 +10,13 @@ import (
 	"github.com/projectdiscovery/subfinder/v2/pkg/subscraping"
 )
 
-// EnumerateSubdomains enumerates all the subdomains for a given domain
-func (a *Agent) EnumerateSubdomains(ctx context.Context, domain string, proxy string, rateLimit, timeout int, maxEnumTime time.Duration) chan subscraping.Result {
+// EnumerateSubdomains wraps EnumerateSubdomainsWithCtx with an empty context
+func (a *Agent) EnumerateSubdomains(domain string, proxy string, rateLimit, timeout int, maxEnumTime time.Duration) chan subscraping.Result {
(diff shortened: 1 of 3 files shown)


tarunKoyalwar:
@EndPositive , How about implementing the same without making any breaking changes to API since many users import subfinder as library . 

Ex:

```go
func (a *Agent) EnumerateSubdomains(domain string, proxy string, rateLimit, timeout int, maxEnumTime time.Duration) chan subscraping.Result {
return a.EnumerateSubdomainsWithCtx(context.Background(),domain , proxy , rateLimit, timeout , maxEnumTime)
}
```

```go
func (a *Agent) EnumerateSubdomainsWithCtx(ctx context.Context, domain string, proxy string, rateLimit, timeout int, maxEnumTime time.Duration) chan subscraping.Result 
```

v2/pkg/passive/passive.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // EnumerateSubdomains enumerates all the subdomains for a given domain
-func (a *Agent) EnumerateSubdomains(domain string, proxy string, rateLimit, timeout int, maxEnumTime time.Duration) chan subscraping.Result {
+func (a *Agent) EnumerateSubdomains(ctx context.Context, domain string, proxy string, rateLimit, timeout int, maxEnumTime time.Duration) chan subscraping.Result {
 	results := make(chan subscraping.Result)
 	go func() {
 		defer close(results)
```
tarunKoyalwar, line 17: @EndPositive , How about implementing the same without making any breaking changes to API since many users import subfinder as library .

tarunKoyalwar (changes requested):

tarunKoyalwar (approved):
lgtm!

tarunKoyalwar:
@ehsandeep , build test is currently failing due to a #761 . It is WIP and for this PR we can either merge it to `dev` or rebase after #761  is resolved

ehsandeep (approved):

PR #769, opened 2023-02-23
Update hunter.go

XTeam-Wing:
The syntax for accurately querying domain names is domain.suffix


tarunKoyalwar:
@XTeam-Wing , can you share what is exact ratelimit of hunter and maybe some reference .
we have a ratelimit [lib](https://github.com/projectdiscovery/ratelimit) for such cases which handles ratelimit and delays efficiently (without hitting 429)

XTeam-Wing:
> @XTeam-Wing , can you share what is exact ratelimit of hunter and maybe some reference . we have a ratelimit [lib](https://github.com/projectdiscovery/ratelimit) for such cases which handles ratelimit and delays efficiently (without hitting 429)

Too fast will be like this.
image

Issue #814, opened 2023-04-22
Hope to better support 360quake


chushuai:
360quake has detected 1.3 billion domain name service data in total, and most of the status codes are 200. I hope that the latest data will be used in the API call process so that the effect will be better！

https://quake.360.net/quake/#/searchResult?searchVal=is_domain%3Atrue&selectIndex=quake_service&latest=true

![image](https://user-images.githubusercontent.com/5875458/233761406-7cd48c59-5d99-4b44-8861-a3bb34089fbe.png)

```
import json

import requests

headers = {
    "X-QuakeToken": "Your API Key",
    "Content-Type": "application/json"
}

def get_subdomain(domain: str):
    ret = []
    query_string = f'domain: *.{domain}'
    post_data = {
        "query": query_string,
        "include": ["service.http.host"],
        # Reduce the amount of data transmitted by the network and improve the response speed
        "size": 100,
        "latest": True,
        "ignore_cache": False,
        # The latest data is the data after deduplication, and it is also relatively stable data at present
    }

    while True:
        print('your query: ', json.dumps(post_data))
        response = requests.post(url="https://quake.360.cn/api/v3/scroll/quake_service", headers=headers,
                                 json=post_data)
        result = response.json()

        for data in result["data"]:
            ret.append(data['service']['http']['host'])
        pagination_id = result["meta"]["pagination_id"]
        if not result["data"]:
            break
        post_data['pagination_id'] = pagination_id

    return ret


if __name__ == '__main__':
    for domain in get_subdomain('projectdiscovery.io'):
        print(f'Found {domain}')

```
![image](https://user-images.githubusercontent.com/5875458/233764286-7b50b7c6-bb7e-404f-b13b-c3297708d5ae.png)

chushuai:
https://github.com/projectdiscovery/subfinder/pull/815

PR #815, opened 2023-04-22
fix 360quake API specify the include field to improve efficiency and …

chushuai:
…use latest data (#814 )

fix 360quake API specify the include field to improve efficiency and use latest data (#814 )


ehsandeep (approved):
Thank you @chushuai for the update.

PR #837, opened 2023-05-15
Update the regex to match domain names case in-sensitive

ShubhamRasal:


tarunKoyalwar (approved):
lgtm !

### Notes (cc : @ehsandeep )
- error from riddler source (cause build failure)
- this PR resolves a rare case  where sources contains capitalized subdomain (ex: SCANME.sh)

ehsandeep (approved):

Issue #847, opened 2023-05-26
[Issue] Logging using incorrect level when verbose is set up


reivaxxavier:
**Describe the bug**
Hello there!

I'm encountering a difficulty related to logging in subfinder, which seems to be connected to how gologger handles label and log level. In the piece of code provided below, it seems that whenever an event that includes a label value is logged, this value gets converted to log level during formatting. Is there a specific reasoning behind this design decision?

To illustrate, when I operate subfinder with verbose log level configuration `gologger.DefaultLogger.SetMaxLevel(levels.LevelVerbose)`, I encounter numerous logs with a log level that my external logging system cannot process.

```
{"level":"dnsdumpster","msg":"rate-limited-proxy-108-177-77-100.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"code.l.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"captive.adservice.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"rate-limited-proxy-108-177-64-100.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"google-proxy-66-249-83-100.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"rate-limited-proxy-108-177-73-0.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"rate-limited-proxy-108-177-73-100.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"colab-sandbox.research.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"tensorboard-staging.cloud.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"mail.google.com","timestamp":"2023-05-25T14:07:57+0000"}
```

**Subfinder version**
2.5.8

**Complete command you used to reproduce this**
I'm not running the CLI but rather using the lib, but the code below should reproduce this bug

```
package main

import (
	"bytes"
	"fmt"
	"io"
	"log"

	"github.com/projectdiscovery/subfinder/v2/pkg/resolve"
	"github.com/projectdiscovery/subfinder/v2/pkg/runner"
[...]

PR #848, opened 2023-05-26
fix(logging): label value should be a logging level

gnuletik:
This PR fixes the usage of the `Label(..)` function in gologger.

When Verbose mode is activated, and gologger is configured to use the JSON formatter, here are some logs printed by subfinder:

```
{"level":"dnsdumpster","msg":"rate-limited-proxy-108-177-77-100.google.com","timestamp":"2023-05-25T14:07:57+0000"}
{"level":"dnsdumpster","msg":"code.l.google.com","timestamp":"2023-05-25T14:07:57+0000"}
```

As you can see, the `level` value is incorrectly defined.
The level value should be one of the one defined by gologger:

https://github.com/projectdiscovery/gologger/blob/1fc5289842cc77b4c81dfef7b775cf3b25f97cdf/gologger.go#L15-L22

So, `dnsdumpster` is not the correct value.

This PR fixes the behavior by using logger fields to define the current fields.


olearycrew:
Thanks for this contribution @gnuletik !

It looks like this PR would fix https://github.com/projectdiscovery/subfinder/issues/847

gnuletik:
My pleasure! Yes it will fix this issue :)

tarunKoyalwar:
hii @gnuletik , thanks for the PR but the original issue/feature seems to be gologger and it is being tracked here https://github.com/projectdiscovery/gologger/issues/38

ref: https://github.com/projectdiscovery/subfinder/issues/847#issuecomment-1564923150

ehsandeep:
Thank you @gnuletik for opening this PR, closing this one as it will be addressed in the golagger project.

gnuletik:
Hi @ehsandeep @tarunKoyalwar

I don't think that this issue should be addressed in gologger's https://github.com/projectdiscovery/gologger/issues/38

As I described in the initial description of this PR, gologger is just allowing its users to override the `level` value in the log fields (which may not be a good practice, especially without enum, but that's another subject).

However, subfinder is explicitly NOT using the label field properly, as it is not respecting gologger's defined labels value:
https://github.com/projectdiscovery/gologger/blob/1fc5289842cc77b4c81dfef7b775cf3b25f97cdf/gologger.go#L15-L22

This PR fixes the subfinder usage of subfinder.

gnuletik:
@ehsandeep can you reconsider this PR please?

I really think that the underlying issue is about the usage of gologger in subfinder.

Thanks!

ehsandeep:
@gnuletik see https://github.com/projectdiscovery/subfinder/issues/847#issuecomment-1569746389

PR #849, opened 2023-05-28
Fix #747 - SecurityTrails double subdomains

### LINKED ISSUE #747: [Issue] DNSTrails showing same subdomains for "x.in.net" and "in.net"
**Issue**
The SecurityTrails API appears to show same subdomains for `bhim.in.net` and `in.net`. 

This is not a problem with subfinder, but with the SecurityTrails API. Probable cause of this issue could be bad parsing of domains from subdomains or using a non-reliable list of public suffixes.

I request the maintainers to verify this issue, and temporarily disable the `securitytrails` source by default unless the issue is fixed by SecurityTrails team.

**Evidence**

image

As can be seen in the above screenshot, `securitytrails` source is returning same subdomains for `in.net` TLD and the domain `bhim.in.net`.

n0kovo:


tarunKoyalwar (approved):
lgtm!

```console
$ ./subfinder -d "in.net" -s securitytrails -silent                        130 ↵
tiffanyjewelry.in.net
airmax2015.in.net
whm.speedbroadbandchennai.in.net
cpanel.legal-metrology.ajaydemo.in.net
1111bvsvdfvs.in.net
icloud.com.findmy.in.net

```

n0kovo:
For some context (was in a bit of a hurry when doing this PR), as far as I'm aware, the ST subdomain API used to return only the subdomain strings, hence the code for appending the domain, but now returns the full subdomain.domain.TLD.

ehsandeep (approved):

PR #864, opened 2023-06-01
fix provider config file not used in library mode

### LINKED ISSUE #854: provider config not used/loaded in library mode
### Current Behavior:

When running the command `subfinder -d target.com`, I obtain 367 subdomains for target.com within a duration of 16 seconds and 164 milliseconds. However, when I utilize `subfinder` as a library, I receive only 120 subdomains for target.com, taking 19 seconds and 610 milliseconds.

Both approaches use the same `provider-config.yaml` file. Here is the code I executed:

```go
func main() {
	runnerInstance, err := runner.NewRunner(&runner.Options{
		Threads:            10,                       // Thread controls the number of threads to use for active enumerations
		Timeout:            30,                       // Timeout is the seconds to wait for sources to respond
		MaxEnumerationTime: 10,                       // MaxEnumerationTime is the maximum amount of time in mins to wait for enumeration
		Resolvers:          resolve.DefaultResolvers, // Use the default list of resolvers by marshaling it to the config
		ResultCallback: func(s *resolve.HostEntry) { // Callback function to execute for available host
			log.Println(s.Host, s.Source)
		},
		ProviderConfig: "provider-config.yaml",
	})

	buf := bytes.Buffer{}
	err = runnerInstance.EnumerateSingleDomain("target.com", []io.Writer{&buf})
	if err != nil {
		log.Fatal(err)
	}

	data, err := io.ReadAll(&buf)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("%s", data)
}
```

### Expected Behavior:
I expected both cases to have the same result. Do you know what is wrong here?

nth347:
## Proposed Changes

- fix provider config file not loaded in library mode
- adds updated example usage in library mode at v2/examples/main.go
- run example as part of build test workflow
- closes #854  #867 

> Refer to the issue https://github.com/projectdiscovery/subfinder/issues/854
> 
> The problem is Subfinder unable to load provider-config.yaml file. The pull request added `ParseUserOptions` function to provide library users to parse their own `Options`. Example:
> 
> ```go
> func main() {
> 	// Parse the command line flags and read config files
> 	options := &runner.Options{
> 		ProviderConfig: "C:\\Users\\nth347\\.config\\subfinder\\provider-config.yaml", // Does not take effect if you don't use ParseUserOptions
> 		Output:         os.Stdout,
> 		Domain:         []string{"target.com"},
> 	}
> 	//options.loadProvidersFrom("C:\\Users\\nth347\\.config\\subfinder\\provider-config.yaml")
> 	parsedOptions := runner.ParseUserOptions(options)
> 
> 	newRunner, err := runner.NewRunner(parsedOptions)
> 	if err != nil {
> 		gologger.Fatal().Msgf("Could not create runner: %s\n", err)
> 	}
> 
> 	err = newRunner.RunEnumeration()
> 	if err != nil {
> 		gologger.Fatal().Msgf("Could not run enumeration: %s\n", err)
> 	}
> }
> ```


nth347:
Hi @tarunKoyalwar and @dogancanbakir,

To allow library users to load their own `provider-config.yaml` file, we need to make the `func (options *Options) loadProvidersFrom(location string)` function public. However, currently, Subfinder does not have an API for loading `config.yaml`. In order to address this, I have decided to introduce a new function called `func ParseUserOptions(options *Options) *Options`. This new function performs nearly the same operations as the existing func `ParseOptions() *Options` function, with the difference being that it returns the Options object for further usage & customization by library users.

v2/pkg/runner/options.go
```diff
@@ -314,3 +314,153 @@ func userHomeDir() string {
 	}
 	return usr.HomeDir
 }
+
+// ParseUserOptions parses the command line flags provided by a user
+func ParseUserOptions(options *Options) *Options {
```
dogancanbakir, line 319: The `ParseOptions` function is geared toward the CLI and should not be available in the API. To address this, a new function called `Init()` -or something different- can be introduced for handling Options. It will perform all the tasks of `ParseOptions` except the CLI operations. We can use this new function in `ParseOptions`, `runner.NewRunner`, or any other desired location.

v2/pkg/runner/options.go
```diff
@@ -314,3 +314,153 @@ func userHomeDir() string {
...
+		if err := migrateToProviderConfig(defaultConfigLocation, defaultProviderConfigLocation); err != nil {
+			gologger.Warning().Msgf("Could not migrate providers from existing config %s to provider config %s: %s\n", defaultConfigLocation, defaultProviderConfigLocation, err)
+		} else {
+			// cleanup the existing config file post migration
+			_ = os.Remove(defaultConfigLocation)
+			gologger.Info().Msgf("Migration successful from %s to %s.\n", defaultConfigLocation, defaultProviderConfigLocation)
+		}
+	}
+
+	//options := &Options{}
```
dogancanbakir, line 336: It's a good idea to eliminate any code that's been commented out and that you're not planning on using.

dogancanbakir (changes requested):
Thanks for the PR, @nth347. I just left some comments for you.

v2/pkg/runner/options.go
nth347: I have eliminated the commented code

nth347:
Hi @dogancanbakir after spending some time debugging, it turns out that the new `Init` function is essentially the same as the `ParseOptions` function. However, I did make a change on line 406. If library users specify their `config.yaml` location, it would be beneficial to merge it into the current Options. For now, library users only need to do something like the following:

```go
func main() {
	options := &runner.Options{
		ResultCallback: func(s *resolve.HostEntry) { // Callback function to execute for available host
			log.Println(s.Host, s.Source)
		},
		ProviderConfig: "/path/to/provider-config.yaml", // Will take effect if you call runner.Init
		Config:         "/path/to/config.yaml", // Will take effect if you call runner.Init
		Domain:         []string{"tesla.com"},
	}

	parsedOptions := runner.Init(options)

	newRunner, err := runner.NewRunner(parsedOptions)
	if err != nil {
		gologger.Fatal().Msgf("Could not create runner: %s\n", err)
	}

	err = newRunner.RunEnumeration()
	if err != nil {
		gologger.Fatal().Msgf("Could not run enumeration: %s\n", err)
	}
}
```

tarunKoyalwar (approved):
@nth347  thanks for flagging the issue and creating PR to fix that

I simplified the logic so there is no need to invoke another function like Init() and the updated example for using subfinder as library is available at https://github.com/nth347/subfinder/blob/02f520a0b1926224b8fbf57501a8112d1a7112ee/v2/examples/main.go#L12-L50

nth347:
Hi @tarunKoyalwar, thank you for your fix! I noticed the problem, but I'm still working on improving my coding skills to address it. Your solution is greatly appreciated, and you did an excellent job!

tarunKoyalwar:
@nth347 , no worries also don't hesitate to raise PR thinking about coding skills / being up to the mark . pdteam will help you with PRs as much as we can , just create a draft PR and tag anyone from pdteam

ehsandeep (approved):

PR #903, opened 2023-07-15
Change home directory recovery method

wall6e:
Would it be possible to use `os.UserHomeDir()` instead of `usr.Current().HomeDir` please ?
 
To my knowledge `os.UserHomeDir()` is the only function which take environment variables into account. On Unix, including macOS, it returns the $HOME environment variable. On Windows, it returns %USERPROFILE%. 
More details can be found in the official documentation:
https://pkg.go.dev/os#UserHomeDir

See the test case below: 
Difference between os.UserHomeDir() and usr.Current().HomeDir

The current state causes problems, for example, when using subfinder in an aws lambda. The home directory retrieved with `usr.Current().HomeDir` is "/home/sbx_user1051", whereas this part of the file system is impracticable because is read-only . As it stands, setting the environment variable $HOME doesn't get around this problem, and it's impossible to use a custom configuration.

See below, the output of the `echo 'HOME='\$HOME; /go/bin/subfinder -d xxxx.com` commands:

AWS lambda commands ouput


ehsandeep:
@wall6e see https://github.com/projectdiscovery/utils/pull/172; this change will be added in all the PD projects via utils library upon next release.

v2/pkg/runner/options.go
```diff
@@ -6,7 +6,6 @@ import (
 	"io"
```
captain-hoity-toity, line 6: I guess we're embracing the monolithic monstrosity movement?

dogancanbakir:
Fixed in #1013, closing this.

PR #922, opened 2023-08-07
Adding Deepinfo Source

CyberSecurityRepo:
I think Deepinfo provides an excellent source for subdomains, surpassing the capabilities of numerous other tools. You can find usage specifics on the [documentation](https://docs.deepinfo.com/docs/subdomain-finder-api) page.


ehsandeep:
@CyberSecurityRepo I see you opend a PR with no code changes, was it expected? also this platform is not open for signup so not sure how users will be able to make use of it.

PR #941, opened 2023-08-16
Cleanup migration code

### LINKED ISSUE #930: Cleanup - Provider migration code
**Describe the bug**
In early versions of subfinder the providers were defined within the config file. At some point they were moved to another file `providers.yaml`, and some glue logic was added to guarantee a smooth migration. Such code might no longer be needed since many versions were deployed, consider removing it:

- https://github.com/projectdiscovery/subfinder/blob/0d90c635f64c99309f889c1bdc8e9580596295cd/v2/pkg/runner/options.go#L82
- https://github.com/projectdiscovery/subfinder/blob/0d90c635f64c99309f889c1bdc8e9580596295cd/v2/pkg/runner/options.go#L230

dogancanbakir:
Closes #930.

### DIFF
--- a/v2/pkg/runner/options.go
+++ b/v2/pkg/runner/options.go
@@ -148,6 +148,7 @@ func ParseOptions() *Options {
 	if exists := fileutil.FileExists(defaultConfigLocation); !exists {
 		if err := createProviderConfigYAML(defaultProviderConfigLocation); err != nil {
 			gologger.Error().Msgf("Could not create provider config file: %s\n", err)
+		}
 	}
 
 	if options.Config != defaultConfigLocation {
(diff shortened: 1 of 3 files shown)


v2/pkg/runner/options.go
```diff
@@ -159,6 +145,10 @@ func ParseOptions() *Options {
 		os.Exit(1)
 	}
 
+	if err := createProviderConfigYAML(defaultProviderConfigLocation); err != nil {
```
Mzack9999, line 148: We need to ensure that the file doesn't exist, otherwise the file with providers containing all the keys will be always overwritten at every run

Mzack9999 (changes requested):

Mzack9999 (approved):
lgtm
```console
$ rm -rf /Users/xxx/.config/subfinder/provider-config.yaml 
$ echo scanme.sh | go run .
...
$ file /Users/xxx/.config/subfinder/provider-config.yaml
/Users/xxx/.config/subfinder/provider-config.yaml: ASCII text
```

PR #942, opened 2023-08-16
Add warning for passivetotal

### LINKED ISSUE #920: PassiveTotal api no longer works
Version v2.6.1

PassiveTotal was bought out by Microsoft and doesn't seem to be the same credentials anymore.

passivetotal

dogancanbakir:
Closes #920.

Note: After reviewing again, we found that the old users can still generate API credentials which they can use for PassiveTotal source, but the same wasn't possible with the new users. I updated the PR by adding a warning to inform users of this.

### DIFF
--- a/v2/pkg/passive/sources.go
+++ b/v2/pkg/passive/sources.go
@@ -47,6 +47,7 @@ import (
 	"github.com/projectdiscovery/subfinder/v2/pkg/subscraping/sources/whoisxmlapi"
 	"github.com/projectdiscovery/subfinder/v2/pkg/subscraping/sources/zoomeye"
 	"github.com/projectdiscovery/subfinder/v2/pkg/subscraping/sources/zoomeyeapi"
+	mapsutil "github.com/projectdiscovery/utils/maps"
 )
 
 var AllSources = [...]subscraping.Source{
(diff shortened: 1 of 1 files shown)


v2/pkg/passive/sources.go
```diff
@@ -93,11 +93,19 @@ var AllSources = [...]subscraping.Source{
 	// &reconcloud.Source{}, // failing due to cloudflare bot protection
 }
 
+var sourceWarnings = map[string]string{
```
Mzack9999, line 96: Maybe we can use a `SyncLockMap` to prevent possible future data races

v2/pkg/passive/sources.go
```diff
@@ -93,11 +93,19 @@ var AllSources = [...]subscraping.Source{
...
+	"passivetotal": "New API credentials for PassiveTotal can't be generated, but existing user account credentials are still functional. Please ensure your integrations are using valid credentials.",
+}
+
 var NameSourceMap = make(map[string]subscraping.Source, len(AllSources))
 
 func init() {
 	for _, currentSource := range AllSources {
 		NameSourceMap[strings.ToLower(currentSource.Name())] = currentSource
+
+		if warning, ok := sourceWarnings[strings.ToLower(currentSource.Name())]; ok {
```
Mzack9999, line 106: Wouldn't this potentially print the warning two times? In my opinion it makes sense to print the warning in case the source is selected for a search

Mzack9999 (changes requested):
suggesting minor improvement

v2/pkg/passive/sources.go
dogancanbakir: Uggh, forgot to remove it from here /shrug

Mzack9999 (approved):
lgtm!

ehsandeep (approved):

PR #891, opened 2023-07-04
add rate limit per source option

dogancanbakir:
With this PR, users now have the capability to configure the rate limit for each individual source. Closes #718.

```console
$ go run . -d hackerone.com -rls "crtsh=10/s,shodan=1/s"

               __    _____           __         
   _______  __/ /_  / __(_)___  ____/ /__  _____
  / ___/ / / / __ \/ /_/ / __ \/ __  / _ \/ ___/
 (__  ) /_/ / /_/ / __/ / / / / /_/ /  __/ /    
/____/\__,_/_.___/_/ /_/_/ /_/\__,_/\___/_/

                projectdiscovery.io

[INF] Current subfinder version v2.6.0 (latest)
[INF] Loading provider config from /home/vscode/.config/subfinder/provider-config.yaml
[INF] Enumerating subdomains for hackerone.com
www.hackerone.com
hackerone.com
links.hackerone.com
go.hackerone.com
mta-sts.managed.hackerone.com
mta-sts.hackerone.com
resources.hackerone.com
support.hackerone.com
info.hackerone.com
events.hackerone.com
gslink.hackerone.com
3d.hackerone.com
a.ns.hackerone.com
api.hackerone.com
docs.hackerone.com
mta-sts.forwarding.hackerone.com
design.hackerone.com
b.ns.hackerone.com
[INF] Found 18 subdomains for hackerone.com in 10 seconds 182 milliseconds
```

PR #1175, opened 2024-02-29
remove` time.sleep` from sources

### LINKED ISSUE #1174: remove `time.sleep` from sources

dogancanbakir:
Closes #1174

### DIFF
--- a/v2/pkg/runner/options.go
+++ b/v2/pkg/runner/options.go
@@ -243,7 +243,6 @@ func (options *Options) preProcessOptions() {
 
 var defaultRateLimits = []string{
 	"github=30/m",
-	// "gitlab=2000/m",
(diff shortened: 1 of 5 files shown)


Mzack9999 (commented):
Just wanted to double check if this update is due to server side loosening of rate limits (for example GH has still the same rate limits of implementation time)

dogancanbakir:
I believe that before https://github.com/projectdiscovery/subfinder/pull/891, we were trying to limit the requests with `time.Sleep`. Since we now have this feature, we no longer need those. So, this PR simply removes them and, if necessary, adds default rate limits for sources.

Mzack9999 (approved):

ehsandeep (approved):

PR #1190, opened 2024-03-13
Added context to crtsh database query to prevent it from blocking forever if the connection was lost

UFeindschiff:
The db.Query() call used prior blocks until results have been received. However, if using an unreliable connection (e.g. routing all traffic through Tor), the connection may break, resulting in db.Query() to block forever and as a result for the enumeration to never finish if crtsh was used as a datasource (even if the maximum enumeration time was reached).

This MR is a simple duct-tape fix to query the database with the enumeration context (which is usually a timeout context), so the enumeration will at least always finish after the enumeration timeout was reached. While not a very elegant solution to the problem (we may still be attempting to query a broken connection until the maximum enumeration time is reached), it at least assures that the enumeration actually finishes eventually.


Mzack9999 (approved):
lgtm - @UFeindschiff Thanks for this PR, the proposed solution is by no means a dutch tape but a very good fit as it uses native methods and perfectly respect the maximum enumeration timeout without altering the existing functionality

UFeindschiff:
@Mzack9999 I meant duct tape solution in a way that a query may still hang until the context is canceled, which usually means that if the db connection is lost during enumeration, an enumeration will always take the maximum enumeration time. A "proper" solution would be to check every n seconds to see whether the database connection is still alive and if not open a new TCP connection. That however would need to be patched in the postgresql driver.

ehsandeep (approved):

PR #1212, opened 2024-04-02
Make host configurable

### LINKED ISSUE #1211: Zoomeye api endpoints have a new tld
### Description
The Zoomeye endpoint used in [v2/pkg/subscraping/sources/zoomeyeapi/zoomeyeapi.go] is no longer valid as can be seen here:
https://www.zoomeye.hk/doc?channel=api

Their official twitter page also points to the updated domain:
https://twitter.com/zoomeye_team?lang=en

https://api.zoomeye.org should be modified to https://api.zoomeye.hk

### Subfinder version:

Current subfinder version v2.6.6 (latest)

dogancanbakir:
Closes  #1211

docs at https://github.com/projectdiscovery/docs/pull/38

### DIFF
--- a/v2/pkg/subscraping/sources/zoomeyeapi/zoomeyeapi.go
+++ b/v2/pkg/subscraping/sources/zoomeyeapi/zoomeyeapi.go
@@ -63,7 +63,7 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 		}
 		var pages = 1
 		for currentPage := 1; currentPage <= pages; currentPage++ {
-			api := fmt.Sprintf("https://%s/domain/search?q=%s&type=1&s=1000&page=%d", host, domain, currentPage)
+			api := fmt.Sprintf("https://api.%s/domain/search?q=%s&type=1&s=1000&page=%d", host, domain, currentPage)
 			resp, err := session.Get(ctx, api, "", headers)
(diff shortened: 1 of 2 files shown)


v2/pkg/subscraping/sources/zoomeyeapi/zoomeyeapi.go
```diff
@@ -47,14 +48,22 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 			return
 		}
 
+		randomApiInfo := strings.Split(randomApiKey, ":")
```
Mzack9999, line 51: I think the api key here has a different format:
image

Mzack9999 (changes requested):

v2/pkg/subscraping/sources/zoomeyeapi/zoomeyeapi.go
dogancanbakir: Now, they've different hosts for different regions. Therefore, we ask users to add the host they're using in the provider config file. For example, https://github.com/projectdiscovery/docs/pull/38/files#diff-52512d3883ec5968f7eb443fe887c80b4563f990c466d14caa54e17d5b144b70R107

Mzack9999 (approved):

ehsandeep (approved):

Issue #1333, opened 2024-08-05
Allow comments in domains file


dogancanbakir:
This pull requests allows the domains file to have comments, prefixed with `#`. It strips everything after `#` character, and then does not process empty lines.

I.E.
```
# ProjectDiscovery
projectdiscovery.org # This is the main site
example.com # This is a demo website
```
Will be processed as
```
projectdiscovery.org
example.com
```

ref: https://github.com/projectdiscovery/subfinder/pull/1215#issue-2228678327

PR #1215, opened 2024-04-05
Allow comments in domains file

RedYetiDev:
This pull requests allows the domains file to have comments, prefixed with `#`. It strips everything after `#` character, and then does not process empty lines. Closes #1333.

I.E.
```
# ProjectDiscovery
projectdiscovery.org # This is the main site
example.com # This is a demo website
```
Will be processed as
```
projectdiscovery.org
example.com
```

PR #1346, opened 2024-08-13
allow comments in input file

dogancanbakir:
Closes https://github.com/projectdiscovery/subfinder/issues/1333 ( based on #1215)


Mzack9999 (approved):
lgtm - Two suggestions:
- What about adding this functionality in utils as it appears that comments might be useful in multiple tools
- If the previous point makes sense, we should cover the functionality with tests

ehsandeep (approved):

PR #1388, opened 2024-09-24
fix: shodan source filter subdomains properly

Gby56:
Hi ! I noticed that the Shodan source was completely falsely reporting subdomains when querying a subdomain.

Shodan returns all subdomains they can find under the parent domain from the query (if you ask `api.example.com`, you get all subdomains for `example.com` regardless.)

Their API response format is
```json
{
  "domain": "example.com",
  "tags": [],
  "subdomains": [],
  "data": [],
  "more": true
}
```

The bug was that we assembled the subdomains with the queried subdomain, making fake subdomains attached to the queried domain.

Query: `test.example.com`

Actual subdomains under example.com:
- test.example.com
- api.example.com
- www.example.com

Current bug output:
- api.test.example.com
- www.test.example.com

Expected:
- test.example.com (itself)

### DIFF
--- a/v2/pkg/runner/stats.go
+++ b/v2/pkg/runner/stats.go
@@ -30,12 +30,14 @@ func printStatistics(stats map[string]subscraping.Statistics) {
 
 	if len(lines) > 0 {
 		gologger.Print().Msgf("\n Source               Duration      Results     Errors\n%s\n", strings.Repeat("─", 56))
-		gologger.Print().Msgf("%s\n", strings.Join(lines, "\n"))
+		gologger.Print().Msg(strings.Join(lines, "\n"))
+		gologger.Print().Msgf("\n")
 	}
 
 	if len(skipped) > 0 {
 		gologger.Print().Msgf("\n The following sources were included but skipped...\n\n")
-		gologger.Print().Msgf("%s\n\n", strings.Join(skipped, "\n"))
+		gologger.Print().Msg(strings.Join(skipped, "\n"))
+		gologger.Print().Msgf("\n\n")
 	}
 }
 
--- a/v2/pkg/subscraping/sources/bevigil/bevigil.go
+++ b/v2/pkg/subscraping/sources/bevigil/bevigil.go
@@ -48,6 +48,7 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 		})
 		if err != nil {
 			results <- subscraping.Result{Source: s.Name(), Type: subscraping.Error, Error: err}
+			s.errors++
 			session.DiscardHTTPResponse(resp)
 			return
 		}
@@ -57,6 +58,7 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 		err = jsoniter.NewDecoder(resp.Body).Decode(&response)
 		if err != nil {
 			results <- subscraping.Result{Source: s.Name(), Type: subscraping.Error, Error: err}
+			s.errors++
 			resp.Body.Close()
 			return
 		}
@@ -69,6 +71,7 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 
 		for _, subdomain := range subdomains {
 			results <- subscraping.Result{Source: s.Name(), Type: subscraping.Subdomain, Value: subdomain}
+			s.results++
 		}
 
 	}()
--- a/v2/pkg/subscraping/sources/binaryedge/binaryedge.go
+++ b/v2/pkg/subscraping/sources/binaryedge/binaryedge.go
@@ -3,6 +3,7 @@ package binaryedge
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"net/url"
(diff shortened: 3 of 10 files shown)


Gby56:
Fixing the tests

dwisiswant0:
Seems like the previous implementation was already correct, right? I don't see any results outside of what I had input. What am I missing here? How are you reproducing this issue? It would be better if you could go ahead and create the issue first.

image

Gby56:
Sure thing, here is the issue with screenshots and clearer explanation. 
I think the screenshots explain perfectly the behavior of the Shodan API, and why there is extract filtering required for subfinder if a subdomain is enumerated, to only get children of the subdomain.

https://github.com/projectdiscovery/subfinder/issues/1389

v2/pkg/subscraping/sources/shodan/shodan.go
```diff
@@ -75,10 +76,13 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 			}
 
 			for _, data := range response.Subdomains {
-				results <- subscraping.Result{
-					Source: s.Name(), Type: subscraping.Subdomain, Value: fmt.Sprintf("%s.%s", data, domain),
+				value := fmt.Sprintf("%s.%s", data, response.Domain)
+				if strings.HasSuffix(value, domain) {
```
Gby56, line 80: @dwisiswant0 I realized, `HasSuffix` might not be the best, I think actually this could act like a `grep`.

For example, if user supplies `api.veracode.com`, this could match `stg-api.veracode.com` which is not strictly speaking, underneath `.api.veracode.com` ? 
Or is it ok to keep it fuzzy ?

dwisiswant0: I was just about to review this.

It seems that we might not even need to use `strings` pkg here. Since the API actually returns both `domain`  `subdomains`, we could just concatenate them directly without any issues. This way, we avoid any perf hit that could come from relying on the `strings` package.

To address your concern; we can put that aside for now, but we should definitely take note of it (https://github.com/projectdiscovery/subfinder/issues/1389#issuecomment-2376145684). No matter what input we provide, Shodan will return the `subdomains` associated with the root `domain`.

Gby56: Ok great, is there filtering done later in the post-processing of all sources in subfinder ?
I would assume if the user gives a subdomain, they only expect children under that ?

dwisiswant0: Wait, I changed my mind.

dwisiswant0, line 80:
```suggestion
				suffix := fmt.Sprintf(".%s", strings.TrimPrefix(domain, "."))
				if strings.HasSuffix(value, suffix) {
```

dwisiswant0 (changes requested):

v2/pkg/subscraping/sources/shodan/shodan.go
dwisiswant0: This should address your concern.

Gby56: I'll debug and validate this, thanks !
I think it might filter out the original api.example.com though, if we filter for .api.example.com, the subdomain cannot be underneath itself.

dwisiswant0: This is getting more complicated than I initially thought, and it seems like this is just a temporary workaround. Moving forward, we could use the [`x/net/publicsuffix`](https://pkg.go.dev/golang.org/x/net/publicsuffix) pkg to properly compare suffixes by stripping out the root domain (eTLD+1) correctly.

Gby56: image
Ok I think another thing that is confusing is that we do have a filtering in place that removes any subdomain that does not fit under the supplied subdomain. Weirdly though, `analysiscenter.veracode.com` appears on my non-fixed version of subfinder...

Gby56: Nevermind, I think `crtsh` is returning `"*.analysiscenter.veracode.com"` and therefore this passes the validation, so you find the supplied domain as a subdomain of itself

Gby56: I think crtsh is an edge case, given the suffix check is done on all sources at the end, I think shodan should NOT include the subdomain itself if it's found in the list of subdomains, so we're ok here :)

Gby56:
I think we're all OK here, thanks for the help @dwisiswant0  ! This should greatly increase quality, the difference is significant, from 70 subdomains on analysiscenter.veracode.com, to 4, the bug was generating quite some false positives

dwisiswant0 (approved):

dogancanbakir (approved):

v2/pkg/subscraping/sources/shodan/shodan.go
dwisiswant0: @dogancanbakir  - at first I also thought that we should align the behavior with the API provider. But it seems reasonable to apply a special case for Shodan API response since the output may not meet expectations without specific treatment to filter it.

Gby56: From my other comment with the screenshot in the debugger, I saw that [there is indeed a filter ](https://github.com/projectdiscovery/subfinder/blob/6537b327867861ec4eb90d07604f8bb79bdce88d/v2/pkg/runner/enumerate.go#L67)at the end of the whole pipeline for collecting all sources, so maybe my modification for shodan is redundant, we could just concatenate parent domain + subdomains and return all subdomains, and let the existing filter do the work, to avoid unnecessary `strings` usage in Shodan ?

dwisiswant0: Oh, I didn't realize that either, my bad, folks.

I'll go ahead and revert the changes and stick to "just concatenate them directly" as I proposed.

dwisiswant0:
Superseded by #1392

Issue #1399, opened 2024-10-01
[Feature Request] Line-based File Scanning


Bundy01:
### Please describe your feature request:

Some sources are limited in requests, and adding an option to read a file (with the `-dL` option) by lines and then quit Subfinder would be a plus.
For ex:
```sh
subfinder -lines 1,50 -nW -dL FILE
```

### Describe the use case of this feature:

At the moment, I manually go down the first 50 lines to the end of the file and so on.
The following command would save me the trouble.
```
subfinder -lines 50,100 -nW -dL FILE
```

Regards.

GeorginaReeder:
Thanks so much for your feature request @Bundy01 , we'll take a look into this!

dogancanbakir:
This is a nice-to-have feature. You can easily do something like `sed -n '5,7p' FILE > subfinder...` instead of another flag.

Mentioning @ehsandeep for the discussion.

Bundy01:
@dogancanbakir, I think the option would be preferable because not everyone knows the command you put in.

Bundy01:
@dogancanbakir: I tested your command by replacing `>` with `|`.
This is the result I expected, so if you don't add this option in Subfinder, you can close this issue.

PR #1400, opened 2024-10-04
Line-based File Scanning

SkobelkinYaroslav:
Closes [#1399](https://github.com/projectdiscovery/subfinder/issues/1399)

### DIFF
--- a/v2/pkg/runner/options.go
+++ b/v2/pkg/runner/options.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/projectdiscovery/chaos-client/pkg/chaos"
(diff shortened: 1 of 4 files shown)


GeorginaReeder:
Thanks for your contribution @SkobelkinYaroslav !

We also have a [Discord server](https://discord.com/invite/projectdiscovery), which you’re more than welcome to join. It's a great place to connect with fellow contributors and stay updated with the latest developments!

Mzack9999 (changes requested):
@SkobelkinYaroslav would it be possible to describe a bit better the use case for this new flag?
There are already plenty of tools handling file input such as `sed`, `awk`. We have a conservative approach on adding new flags and this seems like a functionality that might be needed also with other tools, so potentially worth being implemented as another separate tool.

dogancanbakir:
Closing; reason: https://github.com/projectdiscovery/subfinder/issues/1399#issuecomment-2423765467.

Issue #1368, opened 2024-09-07
Add merklemap?


Barre:
Hi,

I've been working on [merklemap](https://www.merklemap.com/) lately.

Would you be open on me opening a PR to integrate it's api? I think it would be a good fit for this project.

There is no real docs as of now but here's a CLI I quickly hacked together for reference: https://github.com/Barre/merklemap-cli

Thanks.

ehsandeep:
@Barre, nice project!

We're open to adding new sources as long as they add value to the project. In this case, did you find any instances where your project discovered subdomains that weren't being found with the existing sources in Subfinder?

Does your project also include data from sources or methods other than CT Logs, which is already included in [Chaos](https://chaos.projectdiscovery.io) and a few other sources available in Subfinder?

Barre:
It does not have any other source, as this project aims to be more of a CT monitor than a pure data finder.

I'd say that the main benefits would be that results are paginatable, unlike crt.sh, and that due to the project's architecture, it shouldn't be overloaded as easily as crt.sh.

Let me know if you think it adds enough value or not! :)​​​​​​​​​​​​​​​​

UNKNOWN0000N:
PLEASE HELP ON HOW TO LOOK FOR OPEN HOSTS AND HOW TO SCAN FOR OPEN HOSTS
USING TERMUX USING TERMUX COMMANDS

dogancanbakir:
@Barre I see no harm in trying it out. Let's include it and make a quick comparison to assess its value -it shouldn't take too much time. Let me know if you need any help!

Barre:
Thank you!

I'll try to push a PR this week.

UNKNOWN0000N:
HELP ME ON HOW TO LOOK FOR OPEN FREE HOSTS ON TERMUX UDING THE NECESSARY
COMMANDS PLIZZ 😭 FOR VPN USE

PR #1482, opened 2024-12-19
add merklemap

dogancanbakir:
Closess #1368


ehsandeep (changes requested):
Getting cloudflare captcha error for every request, was it supposed to work this way?

```
./subfinder -d hackerone.com -s merklemap -v
...
...
[WRN] Encountered an error with source merklemap: unexpected status code 403 received from https://api.merklemap.com/search?stream=true&stream_progress=true&query=*.hackerone.com&page=0
[INF] Found 0 subdomains for hackerone.com in 108 milliseconds 682 microseconds
```

Bundy01:
The first use of this source worked.
I scanned again today and found the same error as ehsandeep.

On the site, access to the API has become not free :(

dogancanbakir:
There are some updates made on the API side. Mentioning @Barre for the discussion.

Barre:
Hi,

To clarify things: Both an API token and an active subscription are now required to access these APIs.

Bundy01:
It's a bad thing to propose API integration and then to switch this API on a pay-only basis.

Barre:
> It's a bad thing to propose API integration and then to switch this API on a pay-only basis.

I am not sure how you think it's possible to offer such services in a sustainable way, for free.

BTW, I was not aware that someone actually implemented merklemap here.

ehsandeep:
@Barre thanks for sharing update.

PR #1529, opened 2025-02-03
FIX: query for downloading subdomains in netlas.go

thurrsense:
Hello!
There was a small issue in creating query for downloading subdomains from Netlas.io.
Adjusted the query syntax to improve subdomain extraction from Netlas.io. The new query "domain:*.%s AND NOT domain:%s" ensures more accurate results by properly filtering the main domain.


GeorginaReeder:
Thanks for your contribution @thurrsense ! :)

dogancanbakir (approved):

PR #1569, opened 2025-04-10
fix(v2/pkg/subscraping): Fix Hunter data source paging issue

YouChenJun:
Optimized the pagination logic of Hunter scanning source, which can correctly obtain multi page results
-Added requests and processing for pages 2 and beyond

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error reporting for API queries, providing immediate feedback on issues such as invalid credentials or bad requests.
  - Enhanced processing of paginated responses by ensuring the first batch is verified before retrieving additional results.
  - Introduced a brief delay between requests to stabilize data retrieval and reduce potential rate-limit issues.
  
- **New Features**
  - Added pagination handling for API responses, allowing retrieval of multiple pages of results from the Hunter and Quake services.

### DIFF
--- a/v2/pkg/subscraping/sources/hunter/hunter.go
+++ b/v2/pkg/subscraping/sources/hunter/hunter.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	jsoniter "github.com/json-iterator/go"
 	"time"
 
-	jsoniter "github.com/json-iterator/go"
 	"github.com/projectdiscovery/subfinder/v2/pkg/subscraping"
 )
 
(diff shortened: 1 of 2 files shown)


YouChenJun:
Hello, I encountered the following error message while using PR 🔨  Build Test / Lint Test (pull_request)Failing after 4s， This problem doesn't seem to have arisen here
@dogancanbakir

YouChenJun:
@dogancanbakir Hello, I encountered the following error when submitting a PR: 🔨 Build Test / Lint Test (pull_request) Failing after 4s. It seems this issue wasn't caused by my changes

v2/pkg/subscraping/sources/hunter/hunter.go
```diff
@@ -93,7 +131,7 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 					s.results++
 				}
 			}
-			pages = int(response.Data.Total/1000) + 1
+			time.Sleep(time.Millisecond * 500)
```
dogancanbakir, line 134: Why do we need this? There is built-in rate limit mechanism, we can use that instead.

v2/pkg/subscraping/sources/hunter/hunter.go
```diff
@@ -57,10 +57,48 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
...
+				s.results++
+			}
+		}
+		// get all pages
+		pages = int(response.Data.Total/1000) + 1
+		//Perform page turning operation
+		for currentPage := 2; currentPage <= pages; currentPage++ {
 			qbase64 := base64.URLEncoding.EncodeToString([]byte(fmt.Sprintf("domain=\"%s\"", domain)))
-			resp, err := session.SimpleGet(ctx, fmt.Sprintf("https://hunter.qianxin.com/openApi/search?api-key=%s&search=%s&page=1&page_size=100&is_web=3", randomApiKey, qbase64))
+			resp, err := session.SimpleGet(ctx, fmt.Sprintf("https://hunter.qianxin.com/openApi/search?api-key=%s&search=%s&page=%d&page_size=100&is_web=3", randomApiKey, qbase64, currentPage))
```
dogancanbakir, line 101: Why do we need a duplicate code block to achieve the same thing? This is the only missing part.

dogancanbakir (changes requested):
Left some comments for you -thanks for the PR!

v2/pkg/subscraping/sources/hunter/hunter.go
YouChenJun: This delay is for rate control during page flipping, and it would be even better if there was a built-in solution

v2/pkg/subscraping/sources/hunter/hunter.go
YouChenJun: I don't quite understand what you mean. The code for obtaining domain assets from Hunter did not correctly retrieve the current page number, and there was no page flipping in the loop

YouChenJun: The original code will only get the data form the first page.

dogancanbakir: Actually, we're getting the page info from the response! The only thing we need to address is updating the page query parameter. Once we do this, the pagination will work smoothly!

YouChenJun: I understand what you mean, but in the for loop, the page parameter in the ulr is always 1, which means we are constantly getting the parameters of the first page. You can debug it or check the API consumption information of Hunter

dDostalker: I think there may be problems with the original code, why do we go to the api and use page=1 every time? I think this pr is a good solution to this problem, if we have neglected others, please kindly tell us, thank you

YouChenJun:
After carefully reading the code, I saw the code that restricts the speed. I will turn off PR and re PR after modifying the code

PR #1637, opened 2025-08-26
Add jsmon

### LINKED ISSUE #1636: add jsmon
https://github.com/projectdiscovery/subfinder/pull/1617

dogancanbakir:
closes https://github.com/projectdiscovery/subfinder/issues/1636

## Summary by CodeRabbit

- New Features
  - Added jsmon as a new passive source for subdomain discovery, enabled by default.
  - Requires an API key and supports multiple keys; one is chosen automatically during scans.
  - Improves coverage and breadth of results; provides run statistics.
- Tests
  - Updated test expectations to include jsmon in the default and all sources sets.


dogancanbakir:
After thoroughly testing it, we realized that it can be non-deterministic and occasionally doesn't produce great results in some domains. So, we've decided not to include it for now, but we'll definitely keep it in mind for future considerations.

PR #1652, opened 2025-10-01
Return enumerated subdomain results in realtime

mukesh-dream11:
Refer to #1649.

I've not tested the `RemoveWildcard = true` path yet, but the code looks fine to me. I've removed the original `ResultCallback` calls and placed them where they will be called sooner, which is hopefully acceptable.

## Summary by CodeRabbit

- **New Features**
  - Results now stream in real time: discovered subdomains are emitted immediately, and entries from wildcard resolution are surfaced as they’re resolved. This reduces latency and improves interactive workflows and integrations.

- **Refactor**
  - Removed the final consolidated emission at the end of a run. Consumers should handle streaming updates during execution rather than relying on a single end-of-run callback. Existing output files and wildcard result queuing behavior remain unchanged.


dogancanbakir:
Thanks for the PR!

PR #1683, opened 2025-12-16
feat: add merklemap source

nohehf:
This PR adds a source for https://www.merklemap.com/ api (docs: https://www.merklemap.com/documentation/search), which allows to search for subdomains in a database of CT logs.
It can be used as a drop in replacement to crt.sh, to prevent spamming it.

## Summary by CodeRabbit

* **New Features**
  * Added merklemap as a new passive data source.
  * Provides paginated retrieval of results, optional API key configuration, per-run statistics, and clearer error reporting when fetches fail.
  * Returns discovered hostnames with associated metadata.

* **Tests**
  * Updated test expectations to include the new source in default and recursive source lists.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/pkg/passive/passive.go
+++ b/pkg/passive/passive.go
@@ -65,11 +65,15 @@ func (a *Agent) EnumerateSubdomainsWithCtx(ctx context.Context, domain string, p
 		for _, runner := range a.sources {
 			wg.Add(1)
 			go func(source subscraping.Source) {
+				defer wg.Done()
 				ctxWithValue := context.WithValue(ctx, subscraping.CtxSourceArg, source.Name())
 				for resp := range source.Run(ctxWithValue, domain, session) {
-					results <- resp
+					select {
+					case <-ctx.Done():
+						return
+					case results <- resp:
+					}
 				}
-				wg.Done()
 			}(runner)
 		}
 		wg.Wait()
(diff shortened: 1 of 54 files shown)


dogancanbakir:
Thanks for the PR! We previously decided not to include this source https://github.com/projectdiscovery/subfinder/pull/1482. We'd love to hear your thoughts on why you believe it should be added and how it could benefit Subfinder users. Looking forward to your insights!

nohehf:
Hey @dogancanbakir ! 
I read #1482, the cloudflare issue is fixed (I contacted them and they changes the WAF rules + you can see in this PR that I use a fixed user agent, which helped).
I believe that this source is great if you want to use subfinder at scale, on a schedule for instance, and don't want to spam crt.sh (which now seem to block major could providers anyway).
It is indeed a paid only api, but I believe it is worth it for enterprises, that have high volume. We plan to use it in combination with other paid APIs.
I do already use my subfinder fork with merklemap at scale and it's been a great drop in replacement for crt.sh.
Hope this helps clarify the use case, I believe this integration does add value. Lmk if you need further info

dogancanbakir:
@nohehf This helps a lot, thanks!

dogancanbakir:
@nohehf Tests fails, we should add it to the default and recursive lists in `sources_test.go` file

dogancanbakir (changes requested):
minor change

nohehf:
Should be good thanks !

Issue #1547, opened 2025-02-19
[Issue] VirusTotal API code 400


Bundy01:
**Describe the bug**
Virustotal API is broken :(

```console
[DBG] Response for failed request against https://www.virustotal.com/api/v3/domains/bistro.io/subdomains?limit=1000:
{"error": {"code": "BadRequestError", "message": "Limit can't be larger than 40"}}
[WRN] Encountered an error with source virustotal: unexpected status code 400 received from https://www.virustotal.com/api/v3/domains/bistro.io/subdomains?limit=1000
[INF] Found 0 subdomains for bistro.io in 490 milliseconds 884 microseconds

```
**Subfinder version**
Current Version: v2.6.8

**Complete command you used to reproduce this**

**Screenshots**
Add screenshots of the error for a better context.

Bundy01:
```
subfinder -s virustotal -d bistro.io -v
```

PR #1703, opened 2026-01-07
fix: revert virustotal api limit to 1000

### LINKED ISSUE #1562: [Issue] Requests of VirusTotal exceeded
**Describe the bug**
Since cf2b9d7686a0697005a58bbb15c1243d26e53e10b, on a file containing 180 lines, VirusTotal's quota of 500 requests is exceeded after only 70 requests.

Of these 70, only about 20 are effective. The others give me an error.

According to a discussion with support, the problem lies with the source code.

Regards.

**Subfinder version**
Current Version: v2.7.0

**Complete command you used to reproduce this**
subfinder -s virustotal -dL FILE

**Screenshots**
Add screenshots of the error for a better context.
```console
unexpected status code 429 received from https://www.virustotal.com/api/v3/domains/cdninstagram.com/subdomains?limit=40
```

dogancanbakir:
## Summary
- Reverts the VirusTotal API limit parameter from 40 back to 1000
- The limit=40 change in #1548 caused excessive pagination requests, quickly exhausting VirusTotal's 500 request quota

## Test plan
- [ ] Test with VirusTotal API key against a domain with many subdomains
- [ ] Verify reduced number of API requests compared to limit=40

Fixes #1562

## Summary by CodeRabbit

* **New Features**
  * Enhanced VirusTotal subdomain enumeration to retrieve up to 1000 subdomains per request, significantly expanding subdomain discovery capabilities and providing more comprehensive enumeration results for reconnaissance activities.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/pkg/subscraping/sources/virustotal/virustotal.go
+++ b/pkg/subscraping/sources/virustotal/virustotal.go
@@ -56,7 +56,7 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 				return
 			default:
 			}
-			var url = fmt.Sprintf("https://www.virustotal.com/api/v3/domains/%s/subdomains?limit=40", domain)
+			var url = fmt.Sprintf("https://www.virustotal.com/api/v3/domains/%s/subdomains?limit=1000", domain)
 			if cursor != "" {
 				url = fmt.Sprintf("%s&cursor=%s", url, cursor)
 			}


Bundy01:
@dogancanbakir, Have you checked that the https://github.com/projectdiscovery/subfinder/issues/1547 issue does not recur?

dogancanbakir:
@Bundy01 Actually, it does, at least for the free API. With the free API, they limit us to 40 requests, and since we're making more requests to get additional data, we end up using all the subscription quota. For premium accounts, we'll follow the same approach. If I'm correct and billing is per request, then we need to differentiate between these two cases to optimize subscription usage.

Bundy01:
@dogancanbakir I don't think there is a solution for #1562.

I built Subfinder with a **value of 1** to understand the requests count thanks to the [4bba544](https://github.com/projectdiscovery/subfinder/commit/4bba544dfeb4237734f8e3facbdbd67364554d1f) commit.

Each result counts as one request + the original request (e.g. 80 results = 81 requests).
If I set the **value to 40**, for 80 results = 3 requests.
And indeed, if I set a **value of 1000**, the source goes into error.

So if I understand correctly, the Virustotal devs misled me by claiming that the quota exceeded problem came from Subfinder.

What do you think?

dogancanbakir:
Yes, we just discussed this with @knakul853 as well. I guess they don't allow the limit to exceed 40. At least, this is our understanding from the document. So, I'll go ahead and close the issue and this PR. Thanks for your time!

PR #1753, opened 2026-02-26
fix: honor -rl (global rate limit) when -rls is not set

Rhan2020:
Fixes #1434.

The global rate limit flag (`-rl`) was effectively ignored for sources that don't have an explicit per-source entry in `-rls`, causing subfinder to run unthrottled unless `-rls` was provided.

Changes:
- Apply `-rl` to every source by default.
- If `-rls` contains an entry for a source, it overrides the global value.
- Guard against nil custom rate limit configuration.
- Clamp negative `-rl` values to 0.

Tests:
- Add coverage for nil custom rate limit (no panic) and negative global rate limit.

/claim #1434

## Summary by CodeRabbit

* **New Features**
  * Per-source rate limits now inherit a computed global default; per-source overrides continue to be supported.

* **Bug Fixes**
  * Negative global rate limits are clamped to zero; global default is applied only when positive.

* **Tests**
  * Added unit tests validating global-default and per-source override behavior across edge cases.

### DIFF
--- a/pkg/passive/passive.go
+++ b/pkg/passive/passive.go
@@ -86,9 +86,9 @@ func (a *Agent) buildMultiRateLimiter(ctx context.Context, globalRateLimit int, 
 	var multiRateLimiter *ratelimit.MultiLimiter
 	var err error
 
-	global := uint(globalRateLimit)
-	if globalRateLimit < 0 {
-		global = 0
+	var global uint
+	if globalRateLimit > 0 {
+		global = uint(globalRateLimit)
 	}
 
 	for _, source := range a.sources {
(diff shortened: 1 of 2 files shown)


Rhan2020:
Addressed CodeRabbit nits:

- Passive RL: avoid casting negative global rate limit to uint before clamping (more readable).
- Tests: assert non-custom source inherits global limit (hackertarget == 3).
- Tests: update unlimited path test to use MaxUint32 so it actually exercises the unlimited branch.

Pushed in afd4e7d.

Rhan2020:
Good catch — updated the test to use MaxUint32 so it actually exercises the unlimited path (the limiter checks against MaxUint32). Current code in this PR already uses `int(math.MaxUint32)`, so we should be good here.\n\nThanks!

Rhan2020:
Thanks — all three points are addressed on the PR branch already:
- `GlobalRateLimitAppliedWhenSourceNotInCustom` now asserts `mrl.GetLimit("hackertarget") == 3`.
- Global limit conversion is now done after a sign check (`var global uint; if globalRateLimit > 0 { global = uint(globalRateLimit) }`).
- The "unlimited" test now passes `math.MaxUint32` to actually hit the unlimited path.

Let me know if you’d prefer we document the per-source `0` behavior as "inherit default" in CLI/docs; happy to add a note.

PR #1763, opened 2026-03-10
feat(report): add interactive intelligence HTML report with robust filtering

realnits:
Introduce a new HTML report pipeline behind the `-oH/--report` flag without changing existing stdout/file/json behavior.

### Highlights
- add report generation module: `pkg/runner/report.go`
- add triage-focused sections:
  - top suspect hosts (risk-ranked)
  - keyword signal density
  - source reliability view
  - per-domain drilldown tables
- add host risk scoring + tagging heuristics for prioritization
- add interactive UX:
  - global search + per-domain search
  - sortable tables
  - CSV export
  - copy-all subdomains
  - expand/collapse domain panels
- improve filter reliability with unified filter evaluation across:
  - text search
  - local domain search
  - keyword pill filters
- add clickable keyword pills + clear filter action
- expand keyword coverage with practical recon/security terms
- apply polished “indie-dev” template styling while keeping report single-file and dependency-free

### Integration changes
- add `ReportFile` option in runner options
- expose output flag: `-oH, --report`
- initialize report collection only when report output is requested
- collect per-domain sources/stats/timing during enumeration
- render report at end of run

### Safety / compatibility
- no behavior change for users not using `--report`
- existing enumeration and output paths remain intact

### Validation
- `go build ./...`
- `go test ./pkg/runner -count=1`

### How the HTML report lookalike

Screenshot 2026-03-10 at 3 23 42 PM

## Summary by CodeRabbit

* **New Features**
  * Added a CLI flag (--report / -oH) to produce a self-contained HTML report file after scans.
  * Reports collect per-domain metrics (timings, per-source statistics, counts) and aggregate global stats.
  * Interactive HTML includes domain accordions, risk tables, keyword and source insights, charts, filtering, CSV export, and copy-to-clipboard.


dogancanbakir:
@realnits Thanks for raising this PR. Could you tell us a bit what was the use case for this you came across and why you think others will benefit from it?

Mzack9999:
Hi @realnits, thanks for the work here, the report looks great.

I don't think it fits in subfinder core though. subfinder is meant to stay small and just output subdomains (via `-json`) so other tools can build on top of it. A reporting and scoring layer moves fast and is fairly opinionated, so it's a better fit outside the enumerator, and it would add a lot of surface area for us to maintain.

What I'd love to see is this as its own tool that reads subfinder's JSON output, e.g.:

    subfinder -d example.com -json -o results.json
    subfinder-report results.json -o report.html

Happy to help make sure the JSON carries everything you need (sources, timing, per-source stats), and we can link it from the README once it's out.

Going to close this one for now, but thanks again for the contribution.

PR #1767, opened 2026-03-20
fix: implement rate limit (rl/rls) support per source (fixes #1434)

promisingcoder:
## Summary

Fixes two bugs preventing `-rl` (global rate limit) and `-rls` (per-source rate limit) options from working correctly.

### Bug 1: Global rate limit (`-rl`) was ignored

In `buildMultiRateLimiter`, `rl` was initialized to `0` and only set when a source existed in the custom rate limit map. Sources not in the map fell through to the `else` branch and got `math.MaxUint32` (unlimited), completely ignoring the `-rl` flag.

**Fix:** Initialize `rl := uint(globalRateLimit)` so the global limit applies as the default for all sources, with per-source configs overriding it.

### Bug 2: Per-source rate limit duration (`-rls`) was lost

`CustomRateLimit` stored only `uint` (the count), dropping the `time.Duration` from `goflags.RateLimit`. `buildMultiRateLimiter` then hardcoded `time.Second` for all limits. So `-rls sitedossier=8/m` was applied as 8 req/sec instead of 8 req/min — 60x too fast.

**Fix:** Added `RateLimitConfig{MaxCount uint; Duration time.Duration}` to the subscraping package, updated `CustomRateLimit` to use it, stored the duration when loading rate limits in the runner, and used the stored duration in `buildMultiRateLimiter`.

### Files Changed
- `pkg/subscraping/types.go` - Added `RateLimitConfig` struct, updated `CustomRateLimit` to store config with duration
- `pkg/runner/runner.go` - Store both count and duration from goflags when building rate limits
- `pkg/passive/passive.go` - Use global rate limit as default, respect per-source duration

/claim #1434

Closes #1434

### DIFF
--- a/pkg/passive/passive.go
+++ b/pkg/passive/passive.go
@@ -86,13 +86,19 @@ func (a *Agent) buildMultiRateLimiter(ctx context.Context, globalRateLimit int, 
 	var multiRateLimiter *ratelimit.MultiLimiter
 	var err error
 	for _, source := range a.sources {
-		var rl uint
-		if sourceRateLimit, ok := rateLimit.Custom.Get(strings.ToLower(source.Name())); ok {
-			rl = sourceRateLimitOrDefault(uint(globalRateLimit), sourceRateLimit)
+		// Start with the global rate limit; per-source limits override it
+		rl := uint(globalRateLimit)
+		duration := time.Second
+
+		if sourceConfig, ok := rateLimit.Custom.Get(strings.ToLower(source.Name())); ok {
+			if sourceConfig.MaxCount > 0 {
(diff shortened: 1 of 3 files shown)


Mzack9999:
Closing this PR as #1764 has been merged to address this issue.

Your PR had the most elegant and minimal implementation (23 additions, 15 deletions) — initializing `rl := uint(globalRateLimit)` as the default was a clean approach, and the `RateLimitConfig` struct bundling count+duration was a good design. However, it lacked tests and nil safety for the `rateLimit` parameter. The merged PR (#1764) covers the same fixes with comprehensive test coverage and nil guards. Thanks for the contribution!

Issue #1778, opened 2026-04-19
[Issue]  Chaos Source Repeats Domain Twice


meme-lord:
**Describe the bug**
For several domains the Chaos source returns results for `xyz.com` in the form `subdomain.xyz.com.xyz.com` for every subdomain when it should just be `subdomain.xyz.com`.

It's not clear if this is a subfinder issue or a Chaos issue but subfinder does add the domain as a suffix to the data. This doesn't happen with every domain

**Subfinder version**
v2.13.0

**Complete command you used to reproduce this**
Bug happens with these domains:
```
subfinder -s chaos -d hotmail.com
subfinder -s chaos -d swisscom.ch
```
But not with:
```
subfinder -s chaos -d sorcery.ie
```

x-stp:
upstream issue.

ehsandeep:
@meme-lord we did some cleanup and validation at the DB level, so this shouldn’t be an issue anymore. Can you confirm?

x-stp:
@ehsandeep for me upstream seems ok; had some to reseed the deps for it to pull properly from the go lib itself; guess fe is cached.

meme-lord:
> [@meme-lord](https://github.com/meme-lord) we did some cleanup and validation at the DB level, so this shouldn’t be an issue anymore. Can you confirm?

Yeah looks good to me. Thanks!

PR #1782, opened 2026-04-20
fix: chaos repeats apex in fqdn output [#1778]

x-stp:
## Proposed changes

Check if fqdn returned ends with the entry domain before its appended; avoid duplication.

### Proof

see #1778; checkout test. 

## Checklist

- [uhu] Pull request is created against the [dev](https://github.com/projectdiscovery/subfinder/tree/dev) branch
- [def] All checks passed (lint, unit/integration/regression tests etc.) with my changes
- [yep] I have added tests that prove my fix is effective or that my feature works
- [see issue] I have added necessary documentation (if appropriate)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subdomain normalization in the Chaos source so upstream variants are consistently returned as fully qualified domain names, avoiding duplicates or missing domain labels.

* **Tests**
  * Added tests covering subdomain normalization edge cases and validating Chaos source metadata (name, default status, recursion support, and key requirement).

### DIFF
--- a/pkg/subscraping/sources/chaos/chaos.go
+++ b/pkg/subscraping/sources/chaos/chaos.go
@@ -3,7 +3,6 @@ package chaos
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/projectdiscovery/chaos-client/pkg/chaos"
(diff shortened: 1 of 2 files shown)


ehsandeep:
@x-stp it was an issue with the source DB (Chaos). We handled it at the source level, so this shouldn’t happen anymore, no code change needed here.

x-stp:
> @x-stp it was an issue with the source DB (Chaos). We handled it at the source level, so this shouldn’t happen anymore, no code change needed here.

great, thanks for that rapid turnaround.
Cheers.

Issue #41, opened 2018-05-10
Exceptionally large scan - out of memory


codingo:
On a large scan the following was encountered when trying to write out to a file at  the duration of the scan. This should probably be worked on with #40 as the solution to this likely aligns up nicely,

```
fatal error: runtime: out of memory

runtime stack:
runtime.throw(0x6db5a8, 0x16)
	/usr/lib/go-1.10/src/runtime/panic.go:616 +0x81
runtime.sysMap(0xc75d9e0000, 0x197ad0000, 0x456300, 0x8738f8)
	/usr/lib/go-1.10/src/runtime/mem_linux.go:216 +0x20a
runtime.(*mheap).sysAlloc(0x85a7e0, 0x197ad0000, 0x7f51add9e5a8)
	/usr/lib/go-1.10/src/runtime/malloc.go:470 +0xd4
runtime.(*mheap).grow(0x85a7e0, 0xcbd67, 0x0)
	/usr/lib/go-1.10/src/runtime/mheap.go:907 +0x60
runtime.(*mheap).allocSpanLocked(0x85a7e0, 0xcbd67, 0x873908, 0x7f51acef0d88)
	/usr/lib/go-1.10/src/runtime/mheap.go:820 +0x301
runtime.(*mheap).alloc_m(0x85a7e0, 0xcbd67, 0x410100, 0xc41fb8c2ff)
	/usr/lib/go-1.10/src/runtime/mheap.go:686 +0x118
runtime.(*mheap).alloc.func1()
	/usr/lib/go-1.10/src/runtime/mheap.go:753 +0x4d
runtime.(*mheap).alloc(0x85a7e0, 0xcbd67, 0x7f51ac010100, 0x41566c)
	/usr/lib/go-1.10/src/runtime/mheap.go:752 +0x8a
runtime.largeAlloc(0x197ace000, 0x450001, 0x7f51ade8ad90)
	/usr/lib/go-1.10/src/runtime/malloc.go:826 +0x94
runtime.mallocgc.func1()
	/usr/lib/go-1.10/src/runtime/malloc.go:721 +0x46
runtime.systemstack(0xc400000000)
	/usr/lib/go-1.10/src/runtime/asm_amd64.s:409 +0x79
runtime.mstart()
	/usr/lib/go-1.10/src/runtime/proc.go:1175

goroutine 1 [running]:
runtime.systemstack_switch()
[...]

PR #1809, opened 2026-07-14
fix(subscraping): cap untrusted source response bodies with a size limit

adamsjack711-ux:
### Summary
Passive-source HTTP responses were read with unbounded `io.ReadAll(resp.Body)` (and `json.Decoder` / `bufio.Scanner`) across the source implementations. A malicious, compromised, or MITM'd upstream API returning a very large body can drive the process to OOM — the client timeout bounds time, not size (and `NewSession` sets `InsecureSkipVerify: true`).

This caps response bodies centrally at the session layer so every source inherits the limit regardless of how it reads the body.

### Changes
- `pkg/subscraping/agent.go`: added `maxResponseBodySize = 50 * 1024 * 1024` (50MB) and a `LimitResponseBody` helper that wraps `resp.Body` in an `io.LimitReader` while preserving `Close`. Applied in the shared `httpRequestWrapper` (covers `Get` / `SimpleGet` / `Post` / `SimplePost` / `HTTPRequest`).
- `pkg/subscraping/sources/threatcrowd/threatcrowd.go`: applied the same helper to its raw `session.Client.Do` call, which bypasses the wrapper.
- `pkg/subscraping/agent_test.go`: unit tests for the cap (oversized body truncated, small body untouched, nil-safe, Close propagation).

50MB is generous enough that legitimate large subdomain lists are unaffected. No timeout, concurrency, or source logic changed.

### Testing
`go build ./...` and `go test ./pkg/...` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added an optional maximum response-size limit for passive HTTP enumeration.
  * Added `--response-size-read` / `-rsr` configuration, with unlimited reading by default.
  * Invalid negative size limits are rejected.
  * Response bodies remain safely closable when limits are applied.

* **Bug Fixes**
  * Prevented oversized or continuously streamed responses from consuming excessive resources.
  * Applied response-size protection to ThreatCrowd data retrieval.

* **Documentation**
  * Documented the new response-size setting and command-line options.

* **Tests**
  * Added coverage for capped, uncapped, small, nil, and streamed responses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,8 +106,9 @@ DEBUG:
   -ls, -list-sources  list all available sources
 
 OPTIMIZATION:
-  -timeout int   seconds to wait before timing out (default 30)
-  -max-time int  minutes to wait for enumeration results (default 10)
+  -timeout int                  seconds to wait before timing out (default 30)
+  -max-time int                 minutes to wait for enumeration results (default 10)
+  -rsr, -response-size-read int max response body size to read in bytes from passive sources (0 = unlimited)
 ```
 
 ## Environment Variables
--- a/pkg/passive/passive.go
+++ b/pkg/passive/passive.go
@@ -15,8 +15,9 @@ import (
 )
 
 type EnumerationOptions struct {
-	customRateLimiter *subscraping.CustomRateLimit
-	maxResults        int
+	customRateLimiter   *subscraping.CustomRateLimit
+	maxResults          int
+	maxResponseBodySize int64
 }
 
 type EnumerateOption func(opts *EnumerationOptions)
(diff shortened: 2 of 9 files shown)


Mzack9999:
Thanks for the hardening PR. The central LimitReader on the session path is a clean place for this.

Before merging:

1. Did you hit a real OOM from an oversized passive-source response (repro, source, or crash dump), or is this preventive? We only found an old large-scan memory issue (#41), not one tied to unbounded HTTP bodies.

2. Default behavior: subfinder’s goal is to return as many subdomains as sources provide. A hard size cap changes that by truncating large legitimate responses. Prefer keeping current behavior as the default and making a body-size limit optional (flag / config) for operators who want the memory guard.

Mzack9999 (approved):

PR #1810, opened 2026-07-20
feat(sources): honor -max-results across paginating sources

Gerijacki:
### Summary

`-max-results` (`-mr`) is currently honored only by the `virustotal` source. Every other paginating source keeps walking its cursor/pages to the end and burns API quota even when the user explicitly asked for a cap.

This applies the same early-stop pattern already used by `virustotal` to the rest of the paginating sources: once `session.MaxResults` results have been emitted, the source returns instead of requesting the next page.

### Sources updated

`censys`, `certspotter`, `shodan`, `securitytrails`, `quake`, `dnsdb`, `onyphe`, `rsecloud`, `redhuntlabs`, `urlscan`, `merklemap`, `commoncrawl`.

Single-request sources are unaffected (there is nothing to paginate).

### Behavior

- `0` (default) means unlimited — existing behavior is unchanged, paid keys are never silently truncated.
- With a positive value, pagination stops as soon as the limit is reached, bounding both results and request count.

### Tests

Added `httptest`-based coverage for `censys` and `certspotter` proving that the cap stops pagination and bounds the number of requests, following the existing `virustotal` test pattern. Also updated the flag help text and README.

## Summary by CodeRabbit

* **New Features**
  * Added consistent `max-results` enforcement across multiple subdomain discovery sources.
  * Scraping now stops early (including pagination) once the configured maximum is reached, reducing extra requests.
  * A value of `0` continues to mean unlimited results.
* **Documentation**
  * Updated CLI help and README guidance to clarify how limits interact with pagination and early stopping.
* **Tests**
  * Added/expanded tests to verify both unlimited retrieval and early-stop behavior when limits are applied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ CONFIGURATION:
   -nW, -active                  display active subdomains only
   -proxy string                 http proxy to use with subfinder
   -ei, -exclude-ip              exclude IPs from the list of domains
-  -mr, -max-results int         limit the number of results per source (0 = unlimited; honored by paginating sources such as virustotal)
+  -mr, -max-results int         limit the number of results per source (0 = unlimited; honored by paginating sources)
 
 DEBUG:
   -silent             show only subdomains in output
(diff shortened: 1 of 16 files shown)


Gerijacki:
Good catch — updated the onyphe `emit` helper to send inside a `select` that also observes `ctx.Done()`, matching the pattern used by the other sources, so it can't block on a cancelled context. Done in a23ed25.

Mzack9999 (approved):

PR #1811, opened 2026-07-20
fix(extractor): match the target domain literally and reject partial hosts

Gerijacki:
### Summary

`NewSubdomainExtractor` builds its regex by interpolating the raw domain:

```go
regexp.Compile(`(?i)[a-zA-Z0-9\*_.-]+\.` + domain)
```

This has two problems:

1. **Dots in the domain are treated as "any character".** For `example.com` the regex also matches text like `exampleXcom`, producing false positives.
2. **The match is unbounded on the right.** Text such as `foo.example.com.evil.org` yields the bogus subdomain `foo.example.com`, which then passes the runner's `HasSuffix("."+domain)` check and reaches the output.

This affects every source that extracts hosts from free text (`crtsh`, `waybackarchive`, `commoncrawl`, `hackertarget`, `digitorus`, `rapiddns`, `urlscan`, `sitedossier`, ...).

### Fix

- Escape the domain with `regexp.QuoteMeta` so its dots (and any other metacharacters) match literally.
- Drop matches that are only a prefix of a longer hostname, while still accepting the trailing-dot FQDN form (`host.example.com.`).

### Tests

Added table-driven tests for the extractor covering literal dots, right-boundary rejection, wildcards, trailing dots, and case-insensitivity. The package had no extractor test before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subdomain extraction accuracy, including correct handling of domains with special characters.
  * Reduced false positives by tightening hostname boundary matching and rejecting mismatched apex/overlong label patterns.
  * Ensured consistent lowercasing of extracted subdomains, with correct behavior for wildcards and trailing-dot FQDNs.
* **Tests**
  * Added unit tests covering domain matching, case normalization, wildcard preservation, trailing-dot handling, and false-positive scenarios.

### DIFF
--- a/pkg/subscraping/extractor.go
+++ b/pkg/subscraping/extractor.go
@@ -13,7 +13,10 @@ type RegexSubdomainExtractor struct {
 // NewSubdomainExtractor creates a new regular expression to extract
 // subdomains from text based on the given domain.
 func NewSubdomainExtractor(domain string) (*RegexSubdomainExtractor, error) {
-	extractor, err := regexp.Compile(`(?i)[a-zA-Z0-9\*_.-]+\.` + domain)
+	// QuoteMeta escapes the domain so its dots (and any other metacharacters)
+	// are matched literally. Without this, "example.com" also matches text like
+	// "exampleXcom" because "." means "any character" in a regex.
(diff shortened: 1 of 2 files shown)


Gerijacki:
Thanks — wrapped the extracted match in `strings.Clone` so a short subdomain no longer keeps the full response body alive. Done in 5bf5db4.

Mzack9999 (approved):

PR #1829, opened 2026-08-14
fix(crtsh): remove statement_timeout from connection string

### LINKED ISSUE #1828: crt.sh SQL source broken - statement_timeout rejected by PgBouncer
crt.sh runs behind PgBouncer, which rejects `statement_timeout` as an unknown startup parameter. lib/pq sends it as part of the connection handshake, and PgBouncer closes the connection immediately.

This breaks the crtsh SQL path entirely — subfinder gets zero results from this source.

**To reproduce:**

Connect to crt.sh with `statement_timeout` in the connection string:

```
host=crt.sh user=guest dbname=certwatch sslmode=disable statement_timeout=30000
```

PgBouncer returns: `unsupported startup parameter: statement_timeout`

**Expected:** subfinder should set `statement_timeout` via a `SET` query after connecting, not as a startup parameter. `connect_timeout` is fine in the connection string since lib/pq handles it client-side.

This was introduced in PR #1723 (fix for #1560) via commit [`fba1cc3`](https://github.com/projectdiscovery/subfinder/commit/fba1cc3d27fbcc24e2265b2b5a62967ff5a5ea95), which added `statement_timeout` as a connection string parameter. PgBouncer rejects unknown startup parameters, so the SQL path has been broken since that commit.

orelg:
﻿## Summary

crt.sh runs behind PgBouncer, which rejects `statement_timeout` as an
unknown startup parameter. This breaks the crtsh SQL path entirely.

The HTTP fallback is unreliable — sometimes it returns results (slowly),
sometimes crt.sh returns 404 and subfinder gets nothing. Either way, the
SQL path is broken.

The fix removes `statement_timeout` from the connection string and sends
it as a `SET` query after connecting. `connect_timeout` stays in the
connection string since lib/pq handles it client-side.

Also adds `binary_parameters=yes` which PgBouncer requires.

Fixes #1828

## Proof

The SQL path always fails. The HTTP fallback is inconsistent, so there are two scenarios:

**Option 1: Before — the HTTP fallback does not work (0 results):**

```
$ subfinder -d hackerone.com -s crtsh -v
[WRN] Encountered an error with source crtsh: pq: unsupported startup parameter: statement_timeout
[WRN] Encountered an error with source crtsh: unexpected status code 404 received from https://crt.sh/?q=%.hackerone.com&output=json
[INF] Found 0 subdomains for hackerone.com in 15 seconds
```

**Option 2: Before — HTTP fallback works, slower than the SQL path:**

```
$ subfinder -d hackerone.com -s crtsh -v
[WRN] Encountered an error with source crtsh: pq: unsupported startup parameter: statement_timeout
[INF] Found 14 subdomains for hackerone.com in 25 seconds
```

**After (this fix — SQL works):**

```
$ subfinder -d hackerone.com -s crtsh -v
[crtsh] api.hackerone.com
[crtsh] docs.hackerone.com
[crtsh] support.hackerone.com
[crtsh] www.hackerone.com
...
[INF] Found 14 subdomains for hackerone.com in 2 seconds
```

SQL path works. No errors. Consistent results.

## Checklist

- [ ] I have added necessary documentation (if appropriate) - not needed

### DIFF
--- a/pkg/subscraping/sources/crtsh/crtsh.go
+++ b/pkg/subscraping/sources/crtsh/crtsh.go
@@ -56,9 +56,7 @@ func (s *Source) Run(ctx context.Context, domain string, session *subscraping.Se
 }
 
 func (s *Source) getSubdomainsFromSQL(ctx context.Context, domain string, session *subscraping.Session, results chan subscraping.Result) int {
-	// connect_timeout: limits connection establishment time (in seconds)
-	// statement_timeout: limits query execution time (in milliseconds)
-	connStr := fmt.Sprintf("host=crt.sh user=guest dbname=certwatch sslmode=disable binary_parameters=yes connect_timeout=%d statement_timeout=%d", session.Timeout, session.Timeout*1000)
+	connStr := fmt.Sprintf("host=crt.sh user=guest dbname=certwatch sslmode=disable binary_parameters=yes connect_timeout=%d", session.Timeout)
 	db, err := sql.Open("postgres", connStr)
 	if err != nil {
 		results <- subscraping.Result{Source: s.Name(), Type: subscraping.Error, Error: err}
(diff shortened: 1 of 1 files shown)


ehsandeep:
Fixed in https://github.com/projectdiscovery/subfinder/pull/1832

PR #1832, opened 2026-08-20
fix(crtsh): set statement_timeout via SQL query to avoid PgBouncer handshake rejection (#1828)

### LINKED ISSUE #1828: crt.sh SQL source broken - statement_timeout rejected by PgBouncer

gcoinstash-cmd:
### Proposed Changes
- Remove statement_timeout startup parameter from the DSN connection string in crtsh.go to prevent handshake termination by PgBouncer.
- Apply statement timeout via an explicit SET statement_timeout = ... query on the session after connecting.
- Preserves connect_timeout in the DSN string.

Fixes #1828

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling for certificate searches, helping prevent operations from running longer than expected.
  * Added clearer reporting when database connections or requests cannot be completed.
  * Improved failure handling so unsuccessful operations stop safely and predictably.
  * Ensured connections are released after searches, improving service stability and resource usage during repeated requests.


ehsandeep (approved):

gcoinstash-cmd:
Hi @ehsandeep, thanks for reviewing and merging! Could you please trigger the Algora reward command (/reward @gcoinstash-cmd) on this PR/issue when you have a moment? Thank you!

PR #1834, opened 2026-08-22
feat: add scanmalware passive source

jonaslejon:
## Proposed changes

Adds [ScanMalware](https://scanmalware.com) as a default passive source. **No API key**: the
API is anonymous, and a key only raises the rate limit (600 req/min anonymous), so it works on
a fresh install with nothing configured.

The source makes three queries and merges them, because they return different populations:

| Query | Returns |
| --- | --- |
| `/api/v1/ct/dns/{domain}` | subdomains in Certificate Transparency DNS records — the bulk of the results |
| `/api/v1/search/smql?q=domain:*.{domain}` | hosts in the public scan archive: scanned, but may hold no certificate |
| `/api/v1/hosts/{domain}` | hosts a browser resolved and requested while rendering pages on the domain |

The third is the one CT-based sources can't reach. Dev and staging hosts that hold no
certificate and were never scanned directly show up only there, e.g.
`npm-beta-proxy.pkg.github.com`.

A few deliberate choices worth flagging for review:

- **`/api/v1/hosts/` is queried at its default source set only.** It can also return hosts named
  in a page's CSP or extracted statically from script source, which roughly doubles the yield,
  but only about a third of the statically extracted ones were ever actually contacted. The API
  returns hosts grouped by source with a `source_meaning` field so a caller can keep that
  distinction; subfinder's output format can't, so the source sticks to hosts genuinely
  contacted rather than laundering guesses into findings.
- **Truncation is surfaced, not swallowed.** Both endpoints report when they've capped a list;
  the source emits that as an error result instead of returning a partial list as if complete.
- **An explicit `User-Agent` is set.** `session.Get` is used rather than `SimpleGet` so the
  header map overrides the random browser UA, which keeps this source's traffic identifiable
  rather than looking like scraping.

## Proof

`go test ./...` passes, including the updated `pkg/passive` source-list expectations.

Measured against every other source subfinder ships (`-all -es scanmalware` vs `-s scanmalware`):

```
github.com   other sources 41,269   scanmalware 65   unique to scanmalware 17
hkr.se       other sources  2,166   scanmalware 69   unique to scanmalware  4
```

Two caveats I'd rather state than have you find:

- Of the 17 unique hosts on `github.com`, **9 resolve**. The unique tail of any source skews
  toward dormant records, since live hosts are already covered elsewhere. Overall output is
  ~88% resolving on a 40-host sample.
- Some unique results are `mail.*` names that NXDOMAIN. Those come from Certificate
  Transparency: certificates issued for hosts later removed. Same behaviour as other CT-backed
  sources.

Baselines move between runs because several sources rate-limit; `hkr.se` measured 2,166-2,189
across three runs on the same afternoon.

## Checklist

---

**Disclosure:** I run ScanMalware, so this is a vendor-submitted source. Flagging it up front
rather than leaving it to be found in review. Happy to adjust the scope of what it queries, or
drop the third query entirely, if you'd rather a new source stayed narrower.

## Summary by CodeRabbit

* **New Features**
  * Added ScanMalware as a passive subdomain enumeration source.
  * Retrieves subdomains from certificate transparency records, scan archives, and browser-observed hosts.
  * Supports pagination, recursive enumeration, cancellation, error reporting, and result filtering.
  * Available by default without requiring an API key.
  * Continues processing independent data requests when one request fails.
  * Broadens passive discovery coverage across multiple public data sources.


jonaslejon:
The `Lint Test` failure here is pre-existing on `dev` rather than something this PR introduces. The
action runs with `only-new-issues: false`, and the one finding is in `pkg/subscraping/agent.go`,
which this PR does not touch:

```
pkg/subscraping/agent.go:53:3: SA1019: (net/http.Transport).Dial has been deprecated since Go 1.7 (staticcheck)
```

It matters more than a red tick, because `Test Builds` declares `needs: [lint]` — so the build, test,
race and example steps are skipped here and there is no green signal to look at.

I have sent the two-line fix separately in #1835 rather than folding it into this PR. Once
that lands this branch should go green with no changes needed here.

For what it is worth, linting only this PR's package is clean:

```console
$ golangci-lint run ./pkg/subscraping/sources/scanmalware/...
0 issues.
```

ehsandeep (approved):
thank you @jonaslejon

PR #1843, opened 2026-08-28
fix(runner): strip userinfo authentication from domain input URLs

gcoinstash-cmd:
### Summary of Changes
- Strips any leading `user:password@` userinfo from domain inputs when supplied as formatted URLs (e.g. `https://user:pass@example.com/path`).
- Ensures clean downstream normalization for passive enumeration and resolver pipelines.
- Added regression test case in `pkg/runner/util_test.go`.

### Test Validation
- `go test -v ./pkg/runner -run TestPreprocessDomain`: **18/18 scenarios passed 100% green**.
- All runner unit tests passed cleanly.


Mzack9999:
Thanks — subfinder takes domains, not URLs. Userinfo isn’t input we support, and resolver sanitizing is a separate path. Closing.

PR #1844, opened 2026-08-28
test(subscraping): add test case for json-quoted and bracketed subdomain extraction

gcoinstash-cmd:
### Summary of Changes
- Adds explicit test coverage in `pkg/subscraping/extractor_test.go` verifying that `RegexSubdomainExtractor` correctly extracts subdomains embedded in structured JSON payloads and array brackets.
- Validates delimiter handling and prevents false positive boundary truncations.

### Test Validation
- `go test -v ./pkg/subscraping -run TestRegexSubdomainExtractor_Extract`: **10/10 test scenarios passed 100% green**.
- All subscraping package test suites passed cleanly.


Mzack9999:

PR #1845, opened 2026-08-29
test(runner): add custom port URL normalization tests to TestPreprocessDomain

gcoinstash-cmd:
### Summary of Changes
- Adds test cases in `TestPreprocessDomain` for URLs containing custom ports (e.g. `https://example.com:8443/api` and `http://sub.example.com:8080?q=1`).
- Confirms that custom port definitions remain intact while stripping schemes, paths, and query fragments during preprocessing.

### Test Validation
- `go test -v ./pkg/runner -run TestPreprocessDomain`: **20/20 test cases passed 100% green in 0.699s**.


Mzack9999:

PR #1846, opened 2026-08-29
test(subscraping): add url context extraction test case in TestRegexSubdomainExtractor_Extract

gcoinstash-cmd:
### Summary of Changes
- Adds a test case in `TestRegexSubdomainExtractor_Extract` testing extraction of subdomains embedded within complex URL contexts (e.g. `https://auth.example.com/login?redirect=https://admin.example.com:8080`).
- Confirms proper isolation of both apex target subdomains without picking up URL paths, query strings, or port suffixes.

### Test Validation
- `go test -v ./pkg/subscraping -run TestRegexSubdomainExtractor_Extract`: **11/11 test cases passed 100% green in 0.645s**.


Mzack9999:

PR #1856, opened 2026-09-04
feat(shodandomain): add passive source

circleous:
## Proposed changes

Adds `shodandomain` as a keyless passive source reading `https://www.shodan.io/domain/{domain}`. No API key, registered in the default set.

### Why this is not a duplicate of `shodanct`

`shodanct` reads `ctl.shodan.io/api/v1/domain/{d}/hostnames`, which is Shodan's certificate transparency feed. The domain page draws on passive DNS and scan data as well as CT, and neither set contains the other:

| domain | shodanct | shodandomain | in both | only shodanct | only shodandomain |
|---|---|---|---|---|---|
| ycombinator.com | 14 | 49 | 11 | 3 | 38 |
| kernel.org | 76 | 125 | 63 | 13 | 62 |

Checked against crt.sh for `ycombinator.com`: of the 38 names only the domain page returns, just 13 appear in CT. `_dmarc` is the clearest case, since a TXT record can never appear in a certificate SAN. In the other direction `am.mirrors.kernel.org` and `smtp.lkml.kernel.org` hold certificates but are absent from the domain page, so `shodanct` keeps earning its place.

The page returns bare labels, so the source appends the queried domain. `HasRecursiveSupport` is false: a query resolves to the registrable domain, so `/domain/app.example.com` returns `example.com`'s names, which would otherwise be emitted as `www.app.example.com`.

Named after the endpoint to stay distinct from the existing `shodan` (API) and `shodanct` (CT) sources.

### Proof

`-s shodandomain`, keyless: 49 and 125 subdomains for the two domains above, in 94ms on `kernel.org`.

Effect on a keyless default run:

| domain | before | after |
|---|---|---|
| ycombinator.com | 75 | 88 |
| kernel.org | 237 | 244 |

`go build ./...`, `go vet ./...` and `go test ./...` pass, including `TestSourcesWithoutKeys` against `hackerone.com` with no ignore entry needed.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added Shodan domain enumeration as a passive subdomain discovery source.
  - The source is available by name and enabled among the default discovery sources.
  - No API key is required.
  - Discovered subdomains are included alongside results from other configured sources.
  - Supports configured per-source result limits.
  - Retrieval errors are reported without interrupting discovery from other sources.


dogancanbakir:
Thanks for the PR.

A few concerns before we can take this:

1. `https://www.shodan.io/robots.txt` has `Disallow: /domain/` (with `Crawl-delay: 10`). Turning this source on by default would have every subfinder run scrape a path Shodan has blocked, and we don't want to ship that.
2. This page shows the same DNS dataset that the existing `shodan` source already gets from `api.shodan.io/dns/domain`, which is the supported way to access it. For users with a key, this mostly returns the same data twice.
3. Parsing the HTML with a regex will quietly return nothing if the markup changes. Non-200 responses also return without an error.

Because of (1), we're not going to merge this.

Issue #250, opened 2020-06-21
Could not run enumeration (too many open files error)


ZeroDot1:
## What's the problem (or question)?
While performing an automatic scan, the following problem occurred:

`[FTL] Could not run enumeration: open /11/miningpools.cloud.txt: too many open files`
After this message Subfinder stopped working. 
## How can we reproduce the issue?
1. Start Subfinder with the following command, `cat '/URLsToScan.txt' | 'go/bin/subfinder' -config '/Config/subfinder/config.yaml' -t 100 -timeout 1 -silent -oD /11/ | tee /URLsToScanDONEsubfinder.txt`
After 1034 URLS the error occurred.
I don't know if it is a problem in Subfinder or again a problem with my internet connection.

## What are the running context details?

* Installation method (e.g. `pip`, `apt-get`, `git clone` or `zip`/`tar.gz`): `GO111MODULE=on go get -v github.com/projectdiscovery/subfinder/cmd/subfinder`
* Client OS : Linux
* Program version (see banner): 2.3.4
* Exception traceback (if any):
`[FTL] Could not run enumeration: open /11/miningpools.cloud.txt: too many open files`

Mzack9999:
@bauthard - It seems like we need [fdmax](https://github.com/projectdiscovery/fdmax) ?

thealamu:
Is work being done on this? @Mzack9999. May I take a stab at it?

vzamanillo:
We have to check that all the resources used in each source are released correctly before using fdmax, it may not all be released correctly, I think it is a good exercise, there are some golang gotchas for defer, ej. it does not work as expected in iteration blocks.

vzamanillo:
Can we test this against the new 2.4 release?

bauthard:
@ZeroDot1 Can you try to reproduce this with the latest release?

ZeroDot1:
@bauthard Yeah, I'll take another look at that.

ZeroDot1:
@bauthard Yes, I succeeded reproducing the problem...

[FTL] Could not run enumeration: open /11/doipool.com.txt: too many open files

[INF] Current Version: 2.4.1

I use a list of currently 1694 URLs to check for subdomains.

vzamanillo:
Thnka you very much, @ZeroDot1 

[...]

PR #1873, opened 2026-09-09
chore(subscraping): restore HTTP connection reuse

dwisiswant0:
## Proposed changes

Drop the default Connection: close header so follow-up requests to the
same host can reuse a session's connection pool. That skips repeating
TCP setup and TLS handshakes on paginated sources. Explicit Connection
headers from sources still win.

The header landed in b03f0933d16a to stop descriptor exhaustion from
#250. Set IdleConnTimeout to 90s to match Go's default transport and
bound how long idle conns stick around. Existing Session.Close cleanup
still drains the pool when enumeration finishes.

Add HTTP and HTTPS regression tests for reuse, explicit close overrides,
and session cleanup. Add a Linux subprocess test that enumerates 128
domains under a 64-descriptor cap, makes two requests per domain, and
checks that each pair shares one connection. Disable enumeration's
session cleanup via a test overlay and the same test dies on the 30th
domain with "too many open files".

Add a benchmark for successive session requests, including connection
counts and allocations.

### Proof

Median of five runs of 500 requests against a
local server:

                         before       after
    HTTP ns/op           197234       82625
    HTTPS ns/op         2871946       88933
    HTTP allocs/op          145          80
    HTTPS allocs/op         929          85
    connections/run         500           1

These numbers are local repeated requests, not full enumeration
throughput.

## Checklist

- [ ] I have added necessary documentation (if appropriate)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved HTTP connection reuse and lifecycle handling during passive enumeration.
  - Requests now retain reusable connections by default, while idle connections are released after a configured timeout.

- **Tests**
  - Added coverage validating connection reuse, explicit connection closure, and cleanup of idle connections.
  - Added Linux-specific coverage for large domain enumerations to ensure requests complete without connection-related errors.

### DIFF
--- a/pkg/passive/connection_linux_test.go
+++ b/pkg/passive/connection_linux_test.go
@@ -1,0 +1,99 @@
+package passive
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"os/exec"
+	"sync/atomic"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/projectdiscovery/subfinder/v2/pkg/subscraping"
+	"github.com/stretchr/testify/require"
+)
+
+// TestEnumerationReleasesConnections guards against descriptor accumulation
+// across domains. The child process isolates the descriptor limit from other tests.
+func TestEnumerationReleasesConnections(t *testing.T) {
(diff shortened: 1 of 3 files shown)


dwisiswant0:
@coderabbitai create an issue to track this, make it short, one paragraph. must be bug oriented.

Mzack9999 (approved):

